PR #2, opened 2023-01-24
feat: geoip cache and ip localization info included in state.json

asmie:
Currently it's recommended to use geoip providers in the order of their occurence. BIN file with IP-location mappings should be taken by each developer himself. Using online providers for grabbin too much addresses at once results in extremaly long wait time for each request to process.

For testing purposes, sample cache file can be used that was generated from the sample.json file.

There are still many TODO's left as when adding more functionalities to the project current design is not enough.


src/geoip_cache.rs
```diff
@@ -0,0 +1,115 @@
...
+                    remove_entry = true;
+                }
+            }
+        }
+
+        {
+            // Beeing here means that we didn't find the entry in cache.
+            let mut rw_cache = self.cache.write().await;
+
+            if remove_entry {
```
lwojtow, line 86: Can we move locking to inside `if`, then take lock below only if `let Ok(entry)` is true?

src/geoip_cache.rs
```diff
@@ -0,0 +1,115 @@
...
+                rw_cache.entries.remove(&ip);
+            }
+
+            for provider in self.providers.iter() {
+                let entry = provider.lookup(ip).await;
+                if let Ok(entry) = entry {
+                    rw_cache.entries.insert(
+                        ip,
+                        CacheEntry {
```
lwojtow, line 95: I'd clone this entry first, then no need to play with `entries.get ... unwrap` below.

src/geoip_cache.rs
```diff
@@ -0,0 +1,115 @@
...
+                    return Some(rw_cache.entries.get(&ip).unwrap().entry.clone());
+                }
+            }
+        }
+
+        None
+    }
+}
+
+impl GeoCache {
```
lwojtow, line 109: Isn't `Default` better?

src/main.rs
```diff
@@ -6,13 +11,19 @@ use std::{
...
 use spectre::{graph::AGraph, graph::Graph};
+use ziggurat_core_geoip::geoip::GeoIPInfo;
+use ziggurat_core_geoip::providers::ip2loc::Ip2LocationService;
+use ziggurat_core_geoip::providers::ipgeoloc::{BackendProvider, IpGeolocateService};
 
+//TODO(asmie): there is some redundancy here as we have ip in the Node structure and one more time
+// in the GeoIPInfo structure.
 #[derive(Default, Clone, Serialize, Deserialize)]
 pub struct Node {
     ip: String,
```
lwojtow, line 22: Why `String` instead of `IpAddr`?

asmie: We can ask @kylegranger as I've not changed this.
Inside the GeoIPInfo (ziggurat-core) structure ip field is IpAddr type:
https://github.com/runziggurat/ziggurat-core/blob/db38b592ee67de60207e7747d4a09e9aa4ece611/ziggurat-core-geoip/src/geoip.rs#L17

src/geoip_cache.rs
asmie: We could but we need also write lock to remove outdated entry.
There is possible that we'd remove outdated entries only when new data is fetched properly (so move `remove_entry` `if` inside `let Ok(entry)` but it could potentially lead to the situation when there will be many outdated entries kept in the cache.
Personally, I don't believe with such situation and I can move removing entries to the place where we've new data for them (well, in such case we can just modify it). As a tradeoff - some entries can be outdated and still kept in cache but they should not be a big deal.

src/main.rs
kylegranger: I kept the String type from NetworkSummary
https://github.com/runziggurat/ziggurat-core/blob/main/ziggurat-core-crawler/src/summary.rs#L24
At the time, I made it a string when we went from `SocketAddr` (with port number) to just plain ip address (was not aware of `IpAddr` type)
If we use `IpAddr`, it should probably be changed there, too.

src/geoip_cache.rs
asmie: Of course it is :) Thanks, changed 👍

src/main.rs
kylegranger: regarding the redundancy, I had an idea.
```
pub struct GeoInfo {
    /// Country name (long name)
    pub country: Option<String>,
    /// City name
    pub city: Option<String>,
    /// Latitude
    pub latitude: Option<f64>,
    /// Longitude
    pub longitude: Option<f64>,
    /// Timezone of the IP
    pub timezone: Option<String>,
    /// ISP name (unavailable for some providers)
    pub isp: Option<String>,
}

pub struct GeoIPInfo {
    /// IP address
    pub ip: IpAddr,
    /// GeoInfo struct
    pub geo_info: GeoInfo
}
```
And in the state.json, only the GeoInfo gets added to the node.
```
pub struct Node {
    ip: String,
    betweenness: f64,
    closeness: f64,
    num_connections: usize,
    geolocation: Option<GeoInfo>,
}
```

src/geoip_cache.rs
lwojtow: > We could but we need also write lock to remove outdated entry.

Yes, that's what I mean. We would take the lock twice: once before removal (inside if remove_entry), and the second before insert (inside if let Ok(entry)...). My reasoning is because provider.lookup can take a long time, right?

src/geoip_cache.rs
asmie: Done.

src/geoip_cache.rs
asmie: Yea, it could be for some reasons. Changed! :)

lwojtow (approved):

src/main.rs
```diff
@@ -106,41 +154,55 @@ fn write_state(infile: &str, outfile: &str) {
...
 }
 
-fn main() {
+#[tokio::main]
+async fn main() {
     let args: Vec<String> = env::args().collect();
-    if args.len() != 3 {
-        println!("\n\nUsage is: cargo run <in-sample.json> <out-state.json> ");
+
+    // TODO(asmie): could be refactored to use eg. clap crate.
```
Rqnsom, line 170: okay
just a thought - we could have a default geoip-cache.json location defined
throw that default file in `~/ziggurat/utils/crunchy/geoip-cache.json` e.g.

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Add a new provider to the list of providers. The providers will be called in the order they
+    /// are added.
+    pub fn add_provider(&mut self, provider: Box<dyn GeoIPService>) {
+        self.providers.push(provider);
+    }
+
+    /// Load the cache from the file.
+    pub async fn load(&self) -> Result<(), io::Error> {
+        let cache_string = fs::read_to_string(&self.cache_file)?;
```
Rqnsom, line 50: hmhm, we can have a check here to check the cache_file stats to see if the file is older than two weeks - if yes, we can discard it (and delete it) before we even read it here since all IPs will be too old

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+    /// store it into cache.
+    pub async fn lookup(&self, ip: IpAddr) -> Option<GeoIPInfo> {
+        let mut remove_entry = false;
+
+        {
+            let cache = self.cache.read().await;
+            let res = cache.entries.get(&ip);
+            if let Some(entry) = res {
+                // Check if the entry is not too old.
+                if entry.last_updated.elapsed().unwrap() < Duration::from_secs(60 * 60 * 24 * 14) {
```
Rqnsom, line 74: `60 * 60 * 24 * 14`

It could be a const or just followed by a comment: `// two weeks`

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+        fs::write(&self.cache_file, cache_string)
+    }
+
+    /// Function look in cache and if not found, it will call the providers to fetch new data and
+    /// store it into cache.
+    pub async fn lookup(&self, ip: IpAddr) -> Option<GeoIPInfo> {
+        let mut remove_entry = false;
+
+        {
+            let cache = self.cache.read().await;
```
Rqnsom, line 70: This whole block (actually from lines 69-85, it's all cache related) should be a function - something like `self.check_cache(ip).await`

such function should also take care of clearing old entries

src/main.rs
```diff
@@ -70,6 +86,31 @@ fn write_state(infile: &str, outfile: &str) {
...
+    // in the current directory. As IPS will have many configurable factors it should be possible to
+    // set them easily from configuration file.
+    geo_cache.add_provider(Box::new(Ip2LocationService::new(
+        "IP2LOCATION-LITE-DB11.BIN",
+    )));
+
+    //TODO(asmie): enabling providers below can cause program to wait extremely long time for response
+    // when exceeding free rate limit. Currently use it only when don't have IP2LOCATION-LITE-DB11.BIN.
+    // That's why they are added on the further places. In the future there should be some timeout mechanism
+    // for each IP address to avoid waiting too long time for response.
```
Rqnsom, line 104: some long-term ideas:
we could create a fully independent IP info service (cloud IP cache server) which would be running 24/7 on some cloud server

our services - like this one here would ask it via REST API queries for IP address info
the 'cloud cache server' would maintain a database which would be updated periodically from the online providers for the IP addresses it was queried previously - which we can feed for the first time with different crawlers

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        if remove_entry {
+            let mut rw_cache = self.cache.write().await;
+            rw_cache.entries.remove(&ip);
+        }
+
+        for provider in self.providers.iter() {
+            let entry = provider.lookup(ip).await;
+            if let Ok(entry) = entry {
```
Rqnsom, line 89:
```suggestion
            if let Ok(ip_geo_info) = entry {
```

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+                let mut rw_cache = self.cache.write().await;
+                let ret_entry = entry.clone();
+                rw_cache.entries.insert(
+                    ip,
+                    CacheEntry {
+                        last_updated: SystemTime::now(),
+                        entry,
```
Rqnsom, line 96: wouldn't this work without an extra variable:
```suggestion
                rw_cache.entries.insert(
                    ip,
                    CacheEntry {
                        last_updated: SystemTime::now(),
                        entry.clone(),
```

Rqnsom (approved):
LGTM with a few suggestions

testdata/geoip-cache.json
```diff
@@ -0,0 +1 @@
...05:00","isp":null}},"89.58.37.217":{"last_updated":{"secs_since_epoch":1674549596,"nanos_since_epoch":491761814},"entry":{"ip":"89.58.37.217","country":"Germany","city":"Karlsruhe","latitude":49.00471878051758,"longitude":8.38582992553711,"timezone":"+01:00","isp":null}},"202.61.202.126":{"last_updated":{"secs_since_epoch":1674549596,"nanos_since_epoch":467072428},"entry":{"ip":"202.61.202.126","country":"Germany","city":"Karlsruhe","latitude":49.00471878051758,"longitude":8.38582992553711,"timezone":"+01:00","isp":null}},"38.88.125.41":{"last_updated":{"secs_since_epoch":1674549596,"nanos_since_epoch":476904522},"entry":{"ip":"38.88.125.41","country":"Canada","city":"Vancouver","latitude":49.284725189208984,"longitude":-123.11276245117188,"timezone":"-08:00","isp":null}},"37.221.197.204":{"last_updated":{"secs_since_epoch":1674549596,"nanos_since_epoch":489821643},"entry":{"ip":"37.221.197.204","country":"Germany","city":"Karlsruhe","latitude":49.00471878051758,"longitude":8.38582992553711,"timezone":"+01:00","isp":null}},"142.132.144.56":{"last_updated":{"secs_since_epoch":1674549596,"nanos_since_epoch":481610364},"entry":{"ip":"142.132.144.56","country":"Germany","city":"Frankfu
```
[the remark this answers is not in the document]
asmie: https://lite.ip2location.com/database-download
Using above base.

src/geoip_cache.rs
asmie: In fact you're right it could but not exactly that way - field name is needed as follows:
```
rw_cache.entries.insert(
                    ip,
                    CacheEntry {
                        last_updated: SystemTime::now(),
                        entry: entry.clone(),
                    },
                );
```

Otherwise you get:
```
error: expected one of `,`, `:`, or `}`, found `.`
  --> src/geoip_cache.rs:96:30
   |
94 |                     CacheEntry {
   |                     ---------- while parsing this struct
95 |                         last_updated: SystemTime::now(),
96 |                         entry.clone(),
   |                         -    ^ expected one of `,`, `:`, or `}`
```

src/geoip_cache.rs
asmie: It will be moved as configurable value from the crunchy config file. I think that in next (or maybe after next :)) PR.

src/main.rs
asmie: That's generally good idea which could allow slow cache updates respecting limits of particular services. I like it.

However, it depends on the project's future. If it'd be used broadly in our projects, well, there'd be a good reason to do this. If we do not plan increasing accuracy or using it somewhere else I'd put it into TODO list.

Rqnsom: > TODO list.

yeah, this is a project, not a task. let's create another view in our improvement bucket these ideas - so we don't forget about those

src/main.rs
asmie: Change done in ziggurat-core.

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub async fn lookup(&self, ip: IpAddr) -> Option<GeoIPInfo> {
+        let mut remove_entry = false;
+
+        {
+            let cache = self.cache.read().await;
+            let res = cache.entries.get(&ip);
+            if let Some(entry) = res {
+                // Check if the entry is not too old.
+                if entry.last_updated.elapsed().unwrap() < Duration::from_secs(60 * 60 * 24 * 14) {
+                    return Some(entry.entry.clone());
```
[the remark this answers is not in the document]
asmie: Changed.

src/geoip_cache.rs
```diff
@@ -0,0 +1,105 @@
...
+                } else {
+                    remove_entry = true;
+                }
```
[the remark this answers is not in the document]
asmie: Done.

src/main.rs
asmie: Will be in config file in next PRs.

src/geoip_cache.rs
asmie: Done.

src/main.rs
asmie: Agree.

src/geoip_cache.rs
asmie: Done.

src/geoip_cache.rs
asmie: Done.

src/geoip_cache.rs
asmie: I think that removing entries is sufficient for now, in case if we'd like to stick to some old records due to geo limits.

Rqnsom (approved):

PR #3, opened 2023-01-26
feat: configuration module for crunchy and geoip

asmie:
Module will be connected to the rest of crunchy in the next PR.


kylegranger (approved):
LGTM

PR #4, opened 2023-01-27
feat: enable configuration across crunchy

asmie:
Enable configuration abilities across crunchy modules.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,8 +2,7 @@
 name = "ziggurat-crunchy"
 version = "0.1.0"
 edition = "2021"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 description = "P2P network crawler data cruncher for graph metrics"
 readme = "Readme.md"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,7 +2,7 @@
 name = "ziggurat-crunchy"
 version = "0.1.0"
 edition = "2021"
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 description = "P2P network crawler data cruncher for graph metrics"
 readme = "Readme.md"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,12 +5,19 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+description = "P2P network crawler data cruncher for graph metrics"
+readme = "Readme.md"
+
 [dependencies]
+anyhow = "1.0"
+clap = { version = "4.0.29", features = ["derive"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 spectre = "0.5.2"
-ziggurat-core-geoip = { git = "https://github.com/asmie/ziggurat-core", rev = "e82f6ed" }
+toml = "0.6.0"
+ziggurat-core-geoip = { git = "https://github.com/runziggurat/ziggurat-core", rev = "508ddb7" }
 
 [dependencies.tokio]
 version = "1.24"
 features = ["full"]
+
--- a/README.md
+++ b/README.md
@@ -54,6 +54,18 @@ The generated output contains processed data that our renderer can directly use.
 ### Command Line
 
 ```
-cargo run --release testdata/sample.json testdata/state.json testdata/geoip-cache.json
+Usage: ziggurat-crunchy [OPTIONS]
+
+Options:
+  -i, --input-sample <INPUT_SAMPLE>    Input file with sample data to process (overrides input from config file)
+  -o, --out-state <OUT_STATE>          Output file with state of the graph (overrides output from config file)
+  -g, --geocache-file <GEOCACHE_FILE>  Output file with geolocation cache (overrides cache from config file)
+  -c, --config-file <CONFIG_FILE>      Configuration file path (if none defaults will be assumed)
+  -h, --help                           Print help
+  -V, --version                        Print version
 ```
 
+Example:
+```
+cargo run --release -- -i testdata/sample.json -o testdata/state.json -g testdata/geoip-cache.json
+```
--- a/src/geoip_cache.rs
+++ b/src/geoip_cache.rs
@@ -1,11 +1,12 @@
 use serde::{Deserialize, Serialize};
 use std::net::IpAddr;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::SystemTime;
 use std::{collections::HashMap, fs, io, time::Duration};
 use tokio::sync::RwLock;
 
+use crate::config::{GeoIPConfiguration, DEFAULT_KEEP_IN_CACHE_DAYS};
 use ziggurat_core_geoip::geoip::{GeoIPService, GeoInfo};
 
 #[derive(Clone, Serialize, Deserialize)]
@@ -27,15 +28,20 @@ pub struct GeoIPCache {
     cache_file: PathBuf,
     /// Cache entries.
     cache: Arc<RwLock<GeoCache>>,
+    /// How many days to keep the cache entries.
+    keep_in_cache_days: u16,
 }
 
 impl GeoIPCache {
     /// Create a new GeoIP cache.
-    pub fn new(cache_file: &Path) -> Self {
+    pub fn new(config: &GeoIPConfiguration) -> Self {
         Self {
             providers: Vec::new(),
-            cache_file: cache_file.to_owned(),
+            cache_file: config.geocache_file_path.clone(),
             cache: Arc::new(RwLock::new(GeoCache::default())),
+            keep_in_cache_days: config
+                .keep_in_cache_days
+                .unwrap_or(DEFAULT_KEEP_IN_CACHE_DAYS),
         }
     }
 
@@ -92,7 +98,9 @@ impl GeoIPCache {
             let res = cache.entries.get(&ip);
             if let Some(entry) = res {
                 // Check if the entry is not too old.
-                if entry.last_updated.elapsed().unwrap() < Duration::from_secs(60 * 60 * 24 * 14) {
+                if entry.last_updated.elapsed().unwrap()
+                    < Duration::from_secs(60 * 60 * 24 * self.keep_in_cache_days as u64)
+                {
                     return Some(entry.info.clone());
                 }
                 remove_entry = true;
--- a/src/main.rs
+++ b/src/main.rs
@@ -194,14 +194,14 @@ async fn main() {
         .unwrap_or_default();
 
     // Override configuration with command line arguments if provided
-    if arg_conf.input_sample.is_some() {
-        configuration.input_file_path = arg_conf.input_sample;
+    if let Some(input_file) = arg_conf.input_sample {
+        configuration.input_file_path = Some(input_file);
     }
-    if arg_conf.out_state.is_some() {
-        configuration.state_file_path = arg_conf.out_state;
+    if let Some(state_file) = arg_conf.out_state {
+        configuration.state_file_path = Some(state_file);
     }
-    if arg_conf.geocache_file.is_some() {
-        configuration.geoip_config.geocache_file_path = arg_conf.geocache_file.unwrap();
+    if let Some(geocache_file) = arg_conf.geocache_file {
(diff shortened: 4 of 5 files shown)


README.md
```diff
@@ -54,6 +54,18 @@ The generated output contains processed data that our renderer can directly use.
...
-cargo run --release testdata/sample.json testdata/state.json testdata/geoip-cache.json
+Usage: ziggurat-crunchy [OPTIONS]
+
+Options:
+  -i, --input-sample <INPUT_SAMPLE>    Input file with sample data to process (overrides input from config file)
+  -o, --out-state <OUT_STATE>          Output file with state of the graph (overrides output from config file)
+  -g, --geocache-file <GEOCACHE_FILE>  Output file with geolocation cache (overrides cache from config file)
+  -c, --config-file <CONFIG_FILE>      Configuration file path (if none defaults will be assumed)
+  -h, --help                           Print help
+  -V, --version                        Print version
```
kylegranger, line 65: lower-case `-v`

src/geoip_cache.rs
```diff
@@ -27,15 +28,20 @@ pub struct GeoIPCache {
     cache_file: PathBuf,
     /// Cache entries.
     cache: Arc<RwLock<GeoCache>>,
+    /// How many days to keep the cache entries.
+    keep_in_cache_days: u16,
```
kylegranger, line 32: just asking: is there a particular reason that this is a `u16`?  And not u32, or even u64 (where would not need the cast below).

src/main.rs
```diff
@@ -160,24 +179,62 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
-    let args: Vec<String> = env::args().collect();
+    let arg_conf = ArgConfiguration::parse();
+    let mut configuration: CrunchyConfiguration;
 
-    // TODO(asmie): could be refactored to use eg. clap crate.
-    if args.len() != 4 {
-        println!("\n\nUsage is: cargo run <in-sample.json> <out-state.json> <geoip-cache.json> ");
-        println!("E.g.:     \x1b[93mcargo run --release testdata/sample.json testdata/state.json\x1b[0m\n");
-        return;
+    if let Some(config_file_path) = arg_conf.config_file {
```
lwojtow, line 190: This `else.... ::default()` looks like code smell. Can `arg_conf.config_file` be mapped with `unwrap_or_default` at the end?

src/main.rs
```diff
@@ -160,24 +179,62 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
+        configuration.input_file_path = Some(PathBuf::from(input_sample));
+    }
+    if let Some(out_state) = arg_conf.out_state {
+        configuration.state_file_path = Some(PathBuf::from(out_state));
+    }
+    if let Some(geocache_file) = arg_conf.geocache_file {
+        configuration.geoip_config.geocache_file_path = PathBuf::from(geocache_file);
+    }
+
+    if fs::metadata(configuration.input_file_path.as_ref().unwrap()).is_err() {
```
lwojtow, line 208: We don't really need metadata here, just checking if file/dir exists, right? Can this check be done with Path or PathBuf, simultaneously checking for is_file ?

src/main.rs
```diff
@@ -160,24 +179,62 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
+    pub input_sample: Option<String>,
+    /// Output file with state of the graph (overrides output from config file)
+    #[clap(short, long)]
+    pub out_state: Option<String>,
+    /// Output file with geolocation cache (overrides cache from config file)
+    #[clap(short, long)]
+    pub geocache_file: Option<String>,
+    /// Configuration file path (if none defaults will be assumed)
+    #[clap(short, long)]
+    pub config_file: Option<String>,
```
lwojtow, line 237: Clap can build `PathBuf` straight away: https://github.com/lukaszwojtow/caverr/blob/main/caverr-cli/src/args.rs

testdata/config.toml
```diff
@@ -0,0 +1,13 @@
+input_file_path = "testdata/sample.json"
```
kylegranger, line 1: Just my opinion here, but I would not make the input and output files part of the configuration.  I consider configs to be mainly 'static', and orthogonal to inputs and outputs.

kylegranger: I'm open to a third opinion here.

README.md
asmie: That one is pre-generated by the clap itself. I'll check how to change it.

testdata/config.toml
asmie: Currently they are optional in the config file (as well as optional from command line) so there is no need to put them there and use them just from command line. 
Well, I believe that we'll get some day to the point, where crunchy will be run daily by some scripts and I could be easier to run it just passing the config file (if files would be in the same place each run).
But for sure, I'll leave possibility to pass them by hand calling crunchy and setting given by hand overwrites setting read from config file.

kylegranger (approved):

Cargo.toml
```diff
@@ -5,12 +5,19 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+description = "P2P network crawler data cruncher for graph metrics"
+readme = "Readme.md"
+
 [dependencies]
+anyhow = "1.0"
```
Rqnsom, line 12: yes! I would like to use it everywhere eventually

Rqnsom (commented):
Will continue later

README.md
asmie: Well, looked into doc and `-V` cannot be changed. The only way I found is to disable built `version` and write it just like the other options.

src/main.rs
asmie: Changed it - that took me a while to think it out.
Thanks!

src/main.rs
asmie: Changed.

src/main.rs
asmie: Done.

Cargo.toml
asmie: We're using it like here:
https://github.com/runziggurat/crunchy/blob/d29427f6f202ae394e1c32cabfb187691e067585/src/config.rs#L48

That was previous PR where only config.rs file was commited without enabling crunchy to use it  (that's why Cargo was not updated to include deps not used by the app). I've splitted to get smaller PRs.

src/geoip_cache.rs
asmie: I (at least :) ) try to give data most accurate types from the logic point of view - I assumed that the reasonable values for this factor is no more than 365 (max year in cache - that's still long) so it fits in u16. I thought it could be surprising for someone if days count could be so big as u64 (then someone could ask question - why so big to keep values under 365). 
So if there are no other significant factors (like performance - but then there are often no rules :) ) I try to make data models or its parts to follow that accuracy rule.

src/main.rs
```diff
@@ -73,9 +76,9 @@ pub fn load_state(filepath: &str) -> CrunchyState {
 // doing too many things. Especially, when I'd like to add here some other stuff like peer sharing it would
 // be too messy. It should be re-designed and divided into smaller functions with appropriate names and
 // functionalities (like computing graphs, counting factors, geolocalization etc).
-async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
-    let mut geo_cache = GeoIPCache::new(Path::new(cachefile));
-    let response = load_response(infile);
+async fn write_state(config: &CrunchyConfiguration) {
```
Rqnsom, line 79: The whole geo_cache struc could be setup in a separate function

src/main.rs
```diff
@@ -160,24 +179,63 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
-    let args: Vec<String> = env::args().collect();
+    let arg_conf = ArgConfiguration::parse();
+    let mut configuration: CrunchyConfiguration;
 
-    // TODO(asmie): could be refactored to use eg. clap crate.
-    if args.len() != 4 {
-        println!("\n\nUsage is: cargo run <in-sample.json> <out-state.json> <geoip-cache.json> ");
-        println!("E.g.:     \x1b[93mcargo run --release testdata/sample.json testdata/state.json\x1b[0m\n");
-        return;
+    configuration = arg_conf
```
Rqnsom, line 190: the whole configuration setup could be moved to a separate function

Rqnsom (approved):
I had to cherry-pick the PR locally to understand what's going on here 😅
- I think I'll need to review it from scratch when all these TODO comments are taken care of

Cargo.toml
asmie: Until it's not enabled in `main.rs` (as `mod config;`) compiler won't complain. I've not enabled it as it would result in used nowhere code - now it's enabled (as we're using it) so it need to be in Cargo.toml :)

src/main.rs
```diff
@@ -160,24 +179,63 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
+        configuration.state_file_path = arg_conf.out_state;
+    }
+    if arg_conf.geocache_file.is_some() {
+        configuration.geoip_config.geocache_file_path = arg_conf.geocache_file.unwrap();
     }
-    if fs::metadata(&args[1]).is_err() {
-        println!("{}: No such file or directory", &args[1]);
+
+    if !configuration.input_file_path.as_ref().unwrap().is_file() {
+        println!(
```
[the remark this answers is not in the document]
asmie: Done.

src/main.rs
```diff
@@ -160,24 +179,63 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
+    // Override configuration with command line arguments if provided
+    if arg_conf.input_sample.is_some() {
+        configuration.input_file_path = arg_conf.input_sample;
+    }
+    if arg_conf.out_state.is_some() {
+        configuration.state_file_path = arg_conf.out_state;
+    }
+    if arg_conf.geocache_file.is_some() {
+        configuration.geoip_config.geocache_file_path = arg_conf.geocache_file.unwrap();
     }
```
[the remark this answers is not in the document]
asmie: Is there any difference to prefer one approach over the other?

src/main.rs
```diff
@@ -160,24 +179,63 @@ async fn write_state(infile: &str, outfile: &str, cachefile: &str) {
...
     let joutput = serde_json::to_string(&state).unwrap();
-    fs::write(outfile, joutput).unwrap();
+    fs::write(config.state_file_path.as_ref().unwrap(), joutput).unwrap();
 }
 
 #[tokio::main]
 async fn main() {
-    let args: Vec<String> = env::args().collect();
+    let arg_conf = ArgConfiguration::parse();
+    let mut configuration: CrunchyConfiguration;
```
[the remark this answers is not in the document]
asmie: According to: https://doc.rust-lang.org/rust-by-example/variable_bindings/declare.html _declare first_
 should not initialize variable and let it to be done later. So there shouldn't be called _Default_ twice.

But you're right - just to be clean, removed additional assigment and now it looks like:
```
    let mut configuration = arg_conf
        .config_file
        .map(|path| {
            CrunchyConfiguration::new(path.to_str().unwrap())
                .expect("could not load configuration file")
        })
        .unwrap_or_default();
```

src/main.rs
asmie: Totally agree. 
And as stated in the TODO: https://github.com/runziggurat/crunchy/blob/d29427f6f202ae394e1c32cabfb187691e067585/src/main.rs#L72 it should be done at some point of time.

src/main.rs
asmie: Totally agree. 
And as stated in the TODO: https://github.com/runziggurat/crunchy/blob/d29427f6f202ae394e1c32cabfb187691e067585/src/main.rs#L72 it should be done at some point of time.

src/main.rs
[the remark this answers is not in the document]
Rqnsom: Isn't the `if let` use case the most peaceful way to unwrap something without panicking - and it's idiomatic?

let's try to imagine replacing all `if let` statements in our codebase with
```rust
if maybe_have_x.is_some() {
    x = maybe_have_x.unwrap();
    [...]
}
```
it doesn't feel right, no?

src/main.rs
Rqnsom: > During the refactorization (previous comment) that should be taken out from here

but I must be honest, this piece of software was not easy to read - it took me more than I expected to understand what was going on to properly review it

Rqnsom: even the previous PR, which I tried to review, but I couldn't foresee how it would fit the future code

maybe we could have a short description of the code structure that'll be delivered later - so the reviewer could understand all the pieces will fit together

src/main.rs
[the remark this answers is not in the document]
asmie: Done.

asmie:
Probably right after IPS will be ready I'll be resolving left todos.

PR #5, opened 2023-01-31
Node changes

kylegranger:
This PR has two main parts
- compute and assign column position and size fields to Node structure.  This data is used by the renderer to position the node vertically, when many nodes are at a single geo-location (e.g., in a column).  At its core, this process involves sorting the nodes by geo-location.
- moved the node creation into a separate module, so main.rs is a bit less cluttered now.  This refactoring will continue in future revisions.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -25,7 +25,9 @@ pub struct Node {
     ip: String,
     betweenness: f64,
     closeness: f64,
-    num_connections: usize,
+    column_position: u32,
+    column_size: u32,
+    connections: Vec<usize>,
     geolocation: Option<GeoInfo>,
 }
 
@@ -34,10 +36,6 @@ pub struct CrunchyState {
     agraph_length: usize,
     elapsed: f64,
     nodes: Vec<Node>,
-    min_betweenness: f64,
-    max_betweenness: f64,
-    min_closeness: f64,
-    max_closeness: f64,
 }
 
 #[allow(dead_code)]
@@ -72,6 +70,74 @@ pub fn load_state(filepath: &str) -> CrunchyState {
     serde_json::from_str(&jstring).unwrap()
 }
 
+
+pub fn compute_columns(nodes: &mut Vec<Node>) -> HashMap<String,u32> {
+    let mut column_stats: HashMap<String,u32> = HashMap::new();
+    for node in nodes {
+        if let Some(geoinfo) = &node.geolocation {
+            if let Some(latitude) = geoinfo.latitude {
+                if let Some(longitude) = geoinfo.longitude {
+                    let ilatitude: i32 = (latitude * 5.0).floor() as i32;
+                    let ilongitude: i32 = (longitude * 5.0).floor() as i32;
+                    let geostr =  format!("{}:{}", ilatitude, ilongitude);
+                    column_stats.entry(geostr.clone()).and_modify(|count| *count += 1).or_insert(1);
+                    node.column_position = column_stats[&geostr];
+                }
+            }
+        }
+    }
+    column_stats
+}
+
+pub fn set_column_positions(nodes: &mut Vec<Node>, column_stats: &mut HashMap<String,u32>) {
+    for node in nodes {
+        if let Some(geoinfo) = &node.geolocation {
+            if let Some(latitude) = geoinfo.latitude {
+                if let Some(longitude) = geoinfo.longitude {
+                    let ilatitude: i32 = (latitude * 5.0).floor() as i32;
+                    let ilongitude: i32 = (longitude * 5.0).floor() as i32;
+                    let geostr = format!("{}:{}", ilatitude, ilongitude);
+                    if let Some(count) = column_stats.get(&geostr) {
+                        node.column_size = *count;
+                    }
+                }
+            }
+        }
+    }
+}
+
+
+async fn create_nodes(network_summary: &NetworkSummary, geo_cache: &GeoIPCache) -> Vec<Node> {
+    let graph: Graph<usize> = Graph::new();
+    let agraph: &Vec<Vec<usize>> = &network_summary.agraph;
+    let (betweenness, closeness) = graph.compute_betweenness_and_closeness(&agraph);
+    let mut nodes = Vec::with_capacity(agraph.len());
+    for i in 0..agraph.len() {
+        let node: Node = Node {
+            ip: network_summary.node_ips[i].clone(),
+            betweenness: betweenness[i],
+            closeness: closeness[i],
+            column_position: 0,
+            column_size: 0,
+            connections: agraph[i].clone(),
+            geolocation: geo_cache
+                .lookup(
+                    network_summary.node_ips[i]
+                        .parse()
+                        .expect("malformed IP address"),
+                )
+                .await,
+        };
+        nodes.push(node);
+    }
+
+    let mut column_stats = compute_columns(&mut nodes);
+    set_column_positions(&mut nodes, &mut column_stats);
+    nodes
+}
+
+
+
 //TODO(asmie): this NEED to be refactorized as currently it is method-level smell (too long)
 // doing too many things. Especially, when I'd like to add here some other stuff like peer sharing it would
 // be too messy. It should be re-designed and divided into smaller functions with appropriate names and
@@ -79,15 +145,8 @@ pub fn load_state(filepath: &str) -> CrunchyState {
 async fn write_state(config: &CrunchyConfiguration) {
     let mut geo_cache = GeoIPCache::new(&config.geoip_config);
     let response = load_response(config.input_file_path.as_ref().unwrap().to_str().unwrap());
-    let agraph = response.result.agraph;
-    let graph: Graph<usize> = Graph::new();
     let start = Instant::now();
-    let (betweenness, closeness) = graph.compute_betweenness_and_closeness(&agraph);
     let elapsed = start.elapsed();
-    let mut min_betweenness: f64 = 10000.0;
-    let mut max_betweenness: f64 = 0.0;
-    let mut min_closeness: f64 = 10000.0;
-    let mut max_closeness: f64 = 0.0;
 
     let res = geo_cache.load().await;
     if res.is_err() {
@@ -130,48 +189,12 @@ async fn write_state(config: &CrunchyConfiguration) {
         )));
     }
 
-    for between in &betweenness {
-        if *between < min_betweenness {
-            min_betweenness = *between;
-        }
-        if *between > max_betweenness {
-            max_betweenness = *between;
-        }
-    }
-    for close in &closeness {
-        if *close < min_closeness {
-            min_closeness = *close;
-        }
-        if *close > max_closeness {
-            max_closeness = *close;
-        }
-    }
-    let mut nodes = Vec::with_capacity(agraph.len());
-    for i in 0..agraph.len() {
-        let node: Node = Node {
-            ip: response.result.node_ips[i].clone(),
-            betweenness: betweenness[i],
-            closeness: closeness[i],
-            num_connections: agraph[i].len(),
-            geolocation: geo_cache
-                .lookup(
-                    response.result.node_ips[i]
-                        .parse()
-                        .expect("malformed IP address"),
-                )
-                .await,
-        };
-        nodes.push(node);
-    }
+    let nodes = create_nodes(&response.result, &geo_cache).await;
 
     let state = CrunchyState {
-        agraph_length: agraph.len(),
+        agraph_length: response.result.agraph.len(),
         elapsed: elapsed.as_secs_f64(),
-        nodes,
-        min_betweenness,
-        max_betweenness,
-        min_closeness,
-        max_closeness,
+        nodes
     };
 
     // Save all changes done to the cache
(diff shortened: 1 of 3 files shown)


kylegranger:
I am moving the geo-location hashing code in separate function (done twice).

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub closeness: f64,
+    pub column_position: u32,
+    pub column_size: u32,
+    pub connections: Vec<usize>,
+    pub geolocation: Option<GeoInfo>,
+}
+
+fn hash_geo_location(latitude: f64, longitude: f64) -> String {
+    // make unique every 0.2, so multiply by 5, convert to integer
+    let ilatitude: i32 = (latitude * 5.0).floor() as i32;
```
lwojtow, line 22:
```suggestion
    let ilatitude = (latitude * 5.0).floor() as i32;
```

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+    let ilongitude: i32 = (longitude * 5.0).floor() as i32;
+    format!("{ilatitude}:{ilongitude}")
+}
+
+// essentially, we sort the nodes into groups at (nearly) same geo-location
+// we use 0.2 degrees for epsilon in both axes.
+// hash gets created from a string created by two numbers
+// increment each time the same location is found
+pub fn set_column_positions(nodes: &mut Vec<Node>) -> HashMap<String, u32> {
+    let mut column_stats: HashMap<String, u32> = HashMap::new();
```
lwojtow, line 32: Stating the type here (probably) isn't needed.

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+                        node.column_size = *count;
+                    }
+                }
+            }
+        }
+    }
+}
+
+pub async fn create_nodes(
+    agraph: &Vec<Vec<usize>>,
```
lwojtow, line 68: Doesn't clippy complain about this? Could it be:
```suggestion
    agraph: &[Vec<usize>],
```
?

kylegranger: Nah, clippy did not complain here.  But did complain here, where `node_ips` was originally a `&Vec<String>` (now corrected):
```
pub async fn create_nodes(
    agraph: &Vec<Vec<usize>>,
    node_ips: &[String],
    geo_cache: &GeoIPCache,
```
I'll make your change.

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: I know why clippy did not complain... the agraph is an argument to a `spectre` graph method...must be vector, and not slice.  But in reality, it's an AGraph... the correct choice.  
```pub async fn create_nodes(
    agraph: &AGraph,
    node_ips: &[String],
    geo_cache: &GeoIPCache,
```
because
```
pub type AGraph = Vec<Vec<usize>>;
```
done

Rqnsom (commented):
I potentially have one more structural comment, but let me think more about it

src/main.rs
```diff
@@ -12,32 +13,17 @@ use std::{
...
+use crate::nodes::{create_nodes, Node};
 use serde::{Deserialize, Serialize};
-use spectre::{graph::AGraph, graph::Graph};
-use ziggurat_core_geoip::geoip::GeoInfo;
+use spectre::graph::AGraph;
 use ziggurat_core_geoip::providers::ip2loc::Ip2LocationService;
 use ziggurat_core_geoip::providers::ipgeoloc::{BackendProvider, IpGeolocateService};
 
-//TODO(asmie): there is some redundancy here as we have ip in the Node structure and one more time
-// in the GeoIPInfo structure.
```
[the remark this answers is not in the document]
asmie: This should be done - on myside I've changed the structures in zg-core - if @kylegranger is using it (`GeoInfo` instead of `GeoIPInfo` - and I think that he is now :) ) then this comment can be removed as todo is complete.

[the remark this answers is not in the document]
kylegranger: This TODO is definitively done, hence its removal.

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+    column_stats
+}
+
+// do the lookup again, and set node's corresponding final column size
+// this field is the same for all nodes in a given column
+pub fn set_column_sizes(nodes: &mut Vec<Node>, column_stats: &mut HashMap<String, u32>) {
+    for node in nodes {
+        if let Some(geoinfo) = &node.geolocation {
+            if let Some(latitude) = geoinfo.latitude {
+                if let Some(longitude) = geoinfo.longitude {
```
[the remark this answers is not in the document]
asmie: I'll look into this after IPS.
Online providers guarantee to give us both latitude and longitude (https://github.com/grantshandy/ipgeolocate/blob/main/src/lib.rs) of specific location.
Offline provider read those two values separately and give us `Option<>` for each of them:
https://github.com/marirs/rust-ip2location/blob/e9bbb0c2cba3ce197fdc8f5000e1cc3bc9c297a0/src/ip2location/db.rs#L277
which makes it potentially possible to have only one of the values.
Will think about the API to handle such case.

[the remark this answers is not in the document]
Rqnsom: > which makes it potentially possible to have only one of the values.

that shouldn't happen unless it's a bug, no?

I just checked the example ip2location file, and the number of longitude occurrences is the same as the number of latitude occurrences:
```zsh
➜  crunchy git:(main) ✗ grep -o longitude geoip-cache.json | wc -l
    2531
➜  crunchy git:(main) ✗ grep -o latitude geoip-cache.json | wc -l
    2531
```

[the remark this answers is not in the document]
asmie: Yes, the question is if the database contains also both params for each IP. Cache is only subset from database ;)  
But in general, there ought to be returned both lat and long factors - that's why I wrote about only potential possibility and maybe we can just assume default value if one of them is not returned (or even both) as I don't expect it will be often case where location would be unknown. If exact location is not known those bases usually use ISP HQ location (from cache observation ;)) so will change it soon.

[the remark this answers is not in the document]
Rqnsom: > done (for now) -- implemented @Rqnsom's suggestion

we'll see if we this leads to a panic eventually - then we can discuss this again

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+pub async fn create_nodes(
+    agraph: &Vec<Vec<usize>>,
+    node_ips: &[String],
+    geo_cache: &GeoIPCache,
+) -> Vec<Node> {
+    let graph: Graph<usize> = Graph::new();
+    let (betweenness, closeness) = graph.compute_betweenness_and_closeness(agraph);
```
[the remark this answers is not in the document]
kylegranger: done

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+    for node in nodes {
+        if let Some(geoinfo) = &node.geolocation {
+            if let Some(latitude) = geoinfo.latitude {
+                if let Some(longitude) = geoinfo.longitude {
+                    let geostr = hash_geo_location(latitude, longitude);
+                    column_stats
+                        .entry(geostr.clone())
+                        .and_modify(|count| *count += 1)
+                        .or_insert(1);
+                    node.column_position = column_stats[&geostr];
```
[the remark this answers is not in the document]
kylegranger: 👍  nice

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub cell_height: u32,
+    // indices of all connected nodes
+    pub connections: Vec<usize>,
+    // used for latitude, longitude, city, country
+    pub geolocation: Option<GeoInfo>,
+}
+
+fn hash_geo_location(latitude: f64, longitude: f64) -> String {
+    // make unique every 0.2 degrees in both axes, so multiply by 5, convert to integer
+    const DEGREE_RESOLUTION: f64 = 1.0 / 0.2;
```
Rqnsom, line 29: nice!

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+use spectre::graph::{AGraph, Graph};
+
+#[derive(Default, Clone, Serialize, Deserialize)]
+pub struct Node {
+    // the ip address, dotted quad, without port number
+    pub ip: String,
```
Rqnsom, line 12:
```suggestion
    /// the ip address, dotted quad, without port number
    pub ip: String,
```
in Rust, `///` are used for commenting public items (should be applied for all elements here)

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+    node_ips: &[String],
+    geo_cache: &GeoIPCache,
+) -> Vec<Node> {
+    let graph: Graph<usize> = Graph::new();
+    // TODO(asmie/kylegranger): make this an associated function for Graph.
+    // it does not use the Graph per se
+    let (betweenness, closeness) = graph.compute_betweenness_and_closeness(agraph);
+    let mut nodes = Vec::with_capacity(agraph.len());
+    let mut cell_stats: HashMap<String, u32> = HashMap::new();
+    for i in 0..agraph.len() {
```
Rqnsom, line 87: it's much easier to read code which has some logical separation with blank lines
see [more](https://www.baeldung.com/cs/clean-code-formatting)
```suggestion
    let mut cell_stats: HashMap<String, u32> = HashMap::new();

    for i in 0..agraph.len() {
```

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+        let geolocation = geo_cache
+            .lookup(node_ips[i].parse().expect("malformed IP address"))
+            .await;
+        let cell_position = get_cell_position(&mut cell_stats, &geolocation);
+        let node: Node = Node {
```
Rqnsom, line 92:
```suggestion
        let cell_position = get_cell_position(&mut cell_stats, &geolocation);

        let node: Node = Node {
```

src/nodes.rs
```diff
@@ -0,0 +1,93 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+use spectre::graph::Graph;
+
+#[derive(Default, Clone, Serialize, Deserialize)]
+pub struct Node {
+    pub ip: String,
+    pub betweenness: f64,
+    pub closeness: f64,
```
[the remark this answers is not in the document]
kylegranger: done

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+// essentially, we sort the nodes into groups at (nearly) same geo-location
+// we use 0.2 degrees for epsilon in both axes.
+// hash gets created from a string created by two numbers
+// increment each time the same location is found
+pub fn get_cell_position(
+    cell_stats: &mut HashMap<String, u32>,
+    geolocation: &Option<GeoInfo>,
+) -> u32 {
+    if let Some(geoinfo) = geolocation {
+        if let Some(latitude) = geoinfo.latitude {
```
Rqnsom, line 62: this might also work here (should be possible)
```suggestion
    if let Some(GeoInfo { latitude: Some(latitude), ..}) = geolocation {
        // [..]
```

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+            betweenness: betweenness[i],
+            closeness: closeness[i],
+            cell_position,
+            cell_height: 0,
+            connections: agraph[i].clone(),
+            geolocation,
+        };
+        nodes.push(node);
+    }
+    // this second pass is necessary, but quite fast
```
Rqnsom, line 103:
```suggestion
    }
    
    // this second pass is necessary, but quite fast
```

src/nodes.rs
```diff
@@ -0,0 +1,106 @@
...
+    cell_stats: &mut HashMap<String, u32>,
+    geolocation: &Option<GeoInfo>,
+) -> u32 {
+    if let Some(geoinfo) = geolocation {
+        if let Some(latitude) = geoinfo.latitude {
+            let longitude = geoinfo
+                .longitude
+                .expect("a longitude must be set if a latitude is set");
+            let geostr = hash_geo_location(latitude, longitude);
+            return *cell_stats
```
Rqnsom, line 67:
```suggestion
            let geostr = hash_geo_location(latitude, longitude);

            return *cell_stats
```

kylegranger:
I am done with current changes.  Additionally (and per DM convo with Karlo...)
- the `column_*` parameters and structs have been renamed `cell_*`
- `column_size` is now `cell_height`
- the assignment of the `cell_position` is now done at the node instantiation.  This eliminates the need for one pass through the node and geo data, just to assign the position.  A second pass to set to the `cell_height` is still performed (after we know what the height is of each cell)

Rqnsom (approved):
LGTM

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

src/nodes.rs
kylegranger: done

PR #6, opened 2023-02-01
feat: Intelligent Peer Sharing

asmie:
Intelligent Peer Sharing (IPS) module - first approach beeing a complete algorithm which takes into account:
- node params (degree, betweenness, eigenvector, closeness)
- node location
- peerlist size
- hot and cold nodes elimination - partly, trying to avoid creating hot nodes by selecting nodes with lower betweenness

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -43,15 +43,15 @@ pub struct GeoIPConfiguration {
 /// Multi-criteria analysis weights
 #[derive(Debug, Clone, Deserialize)]
 pub struct MultiCriteriaAnalysisWeights {
-    /// Weight (importance) of the location factor (used in multi-criteria analysis)
+    /// Weight (importance) of the location factor
     pub location: f64,
-    /// Weight (importance) of the degree factor (used in multi-criteria analysis)
+    /// Weight (importance) of the degree factor
     pub degree: f64,
-    /// Weight (importance) of the eigenvector factor (used in multi-criteria analysis)
+    /// Weight (importance) of the eigenvector factor
     pub eigenvector: f64,
-    /// Weight (importance) of the betweenness factor (used in multi-criteria analysis)
+    /// Weight (importance) of the betweenness factor
     pub betweenness: f64,
-    /// Weight (importance) of the closeness factor (used in multi-criteria analysis)
+    /// Weight (importance) of the closeness factor
     pub closeness: f64,
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -55,18 +55,23 @@ pub struct MultiCriteriaAnalysisWeights {
     pub closeness: f64,
 }
 
+/// GeoLocationMode enum
+#[derive(Debug, PartialEq, Clone, Deserialize)]
+pub enum GeoLocationMode {
+    Off,
+    PreferCloser,
+    PreferDistant,
+}
+
 /// Configuration for Intelligent Peer Sharing module
 #[derive(Debug, Clone, Deserialize)]
 pub struct IPSConfiguration {
     /// Path where peer list file will be written
     pub peer_file_path: Option<PathBuf>,
-    /// Indicates if configuration should be taken into account
-    pub use_geolocation: bool,
-    /// True means we should prefer closer peers, false means we should prefer farther peers
-    pub use_closer_geolocation: bool,
-    /// If pref_location_closer is true, this is the maximum distance in kilometers we should prefer
-    /// closer peers. If pref_location_closer is false, this is the minimum distance in kilometers we
-    /// should prefer farther peers.
+    /// Indicates if configuration should be taken into account and if so what should be
+    /// preferred (closer or distant).
+    pub geolocation: GeoLocationMode,
+    /// This is the max (or min) distance in km between peers
     pub geolocation_minmax_distance_km: u32,
     /// Indicates how many peers must be changed for each node
     pub change_at_least: u32,
@@ -114,8 +119,7 @@ impl Default for IPSConfiguration {
     fn default() -> IPSConfiguration {
         IPSConfiguration {
             peer_file_path: Some(PathBuf::from("testdata/peers.json")),
-            use_geolocation: true,
-            use_closer_geolocation: true,
+            geolocation: GeoLocationMode::PreferCloser,
             geolocation_minmax_distance_km: 1000,
             change_at_least: 1,
             change_no_more: 2,
--- a/src/config.rs
+++ b/src/config.rs
@@ -41,8 +41,33 @@ pub struct GeoIPConfiguration {
 }
 
 /// Configuration for Intelligent Peer Sharing module
-#[derive(Default, Debug, Clone, Deserialize)]
-pub struct IPSConfiguration {} //TODO(asmie): implement with IPS
+#[derive(Debug, Clone, Deserialize)]
+pub struct IPSConfiguration {
+    /// Path where peer list file will be written
+    pub peer_file_path: Option<PathBuf>,
+    /// Indicates if configuration should be taken into account
+    pub pref_location: bool,
+    /// True means we should prefer closer peers, false means we should prefer farther peers
+    pub pref_location_closer: bool,
+    /// If pref_location_closer is true, this is the maximum distance in kilometers we should prefer
+    /// closer peers. If pref_location_closer is false, this is the minimum distance in kilometers we
+    /// should prefer farther peers.
+    pub pref_location_distance: u32,
+    /// Indicates how many peers must be changed for each node
+    pub change_at_least: u32,
+    /// Indicates maximum peers should be changed for each node
+    pub change_no_more: u32,
+    /// Weight (importance) of the location factor (used in multi-criteria analysis)
+    pub location_weight: f64,
+    /// Weight (importance) of the degree factor (used in multi-criteria analysis)
+    pub degree_weight: f64,
+    /// Weight (importance) of the eigenvector factor (used in multi-criteria analysis)
+    pub eigenvector_weight: f64,
+    /// Weight (importance) of the betweenness factor (used in multi-criteria analysis)
+    pub betweenness_weight: f64,
+    /// Weight (importance) of the closeness factor (used in multi-criteria analysis)
+    pub closeness_weight: f64,
+}
 
 impl CrunchyConfiguration {
     pub fn new(conf_path: &str) -> Result<CrunchyConfiguration> {
@@ -74,6 +99,24 @@ impl Default for GeoIPConfiguration {
             ipapico_api_key: Some(String::from("")),
             ipapicom_enable: true,
             ipapicom_api_key: Some(String::from("")),
+        }
+    }
+}
+
+impl Default for IPSConfiguration {
+    fn default() -> IPSConfiguration {
+        IPSConfiguration {
+            peer_file_path: Some(PathBuf::from("testdata/peers.json")),
+            pref_location: true,
+            pref_location_closer: true,
+            pref_location_distance: 1000,
+            change_at_least: 1,
+            change_no_more: 2,
+            location_weight: 0.1,
+            degree_weight: 0.25,
+            eigenvector_weight: 0.25,
+            betweenness_weight: 0.25,
+            closeness_weight: 0.15,
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -46,13 +46,13 @@ pub struct IPSConfiguration {
     /// Path where peer list file will be written
     pub peer_file_path: Option<PathBuf>,
     /// Indicates if configuration should be taken into account
-    pub pref_location: bool,
+    pub use_geolocation: bool,
     /// True means we should prefer closer peers, false means we should prefer farther peers
-    pub pref_location_closer: bool,
+    pub use_closer_geolocation: bool,
     /// If pref_location_closer is true, this is the maximum distance in kilometers we should prefer
     /// closer peers. If pref_location_closer is false, this is the minimum distance in kilometers we
     /// should prefer farther peers.
-    pub pref_location_distance: u32,
+    pub geolocation_minmax_distance_km: u32,
     /// Indicates how many peers must be changed for each node
     pub change_at_least: u32,
     /// Indicates maximum peers should be changed for each node
@@ -107,9 +107,9 @@ impl Default for IPSConfiguration {
     fn default() -> IPSConfiguration {
         IPSConfiguration {
             peer_file_path: Some(PathBuf::from("testdata/peers.json")),
-            pref_location: true,
-            pref_location_closer: true,
-            pref_location_distance: 1000,
+            use_geolocation: true,
+            use_closer_geolocation: true,
+            geolocation_minmax_distance_km: 1000,
             change_at_least: 1,
             change_no_more: 2,
             location_weight: 0.1,
--- a/src/config.rs
+++ b/src/config.rs
@@ -40,6 +40,21 @@ pub struct GeoIPConfiguration {
     pub ipapicom_api_key: Option<String>,
 }
 
+/// Multi-criteria analysis weights
+#[derive(Debug, Clone, Deserialize)]
+pub struct MultiCriteriaAnalysisWeights {
+    /// Weight (importance) of the location factor (used in multi-criteria analysis)
+    pub location: f64,
+    /// Weight (importance) of the degree factor (used in multi-criteria analysis)
+    pub degree: f64,
+    /// Weight (importance) of the eigenvector factor (used in multi-criteria analysis)
+    pub eigenvector: f64,
+    /// Weight (importance) of the betweenness factor (used in multi-criteria analysis)
+    pub betweenness: f64,
+    /// Weight (importance) of the closeness factor (used in multi-criteria analysis)
+    pub closeness: f64,
+}
+
 /// Configuration for Intelligent Peer Sharing module
 #[derive(Debug, Clone, Deserialize)]
 pub struct IPSConfiguration {
@@ -57,22 +72,15 @@ pub struct IPSConfiguration {
     pub change_at_least: u32,
     /// Indicates maximum peers should be changed for each node
     pub change_no_more: u32,
-    /// Weight (importance) of the location factor (used in multi-criteria analysis)
-    pub location_weight: f64,
-    /// Weight (importance) of the degree factor (used in multi-criteria analysis)
-    pub degree_weight: f64,
-    /// Weight (importance) of the eigenvector factor (used in multi-criteria analysis)
-    pub eigenvector_weight: f64,
-    /// Weight (importance) of the betweenness factor (used in multi-criteria analysis)
-    pub betweenness_weight: f64,
-    /// Weight (importance) of the closeness factor (used in multi-criteria analysis)
-    pub closeness_weight: f64,
+    /// Multi-criteria analysis weights
+    pub mcda_weights: MultiCriteriaAnalysisWeights,
 }
 
 impl CrunchyConfiguration {
     pub fn new(conf_path: &str) -> Result<CrunchyConfiguration> {
         let config_string = fs::read_to_string(conf_path)?;
         let crunchy_config: CrunchyConfiguration = toml::from_str(&config_string)?;
+        println!("{:?}", crunchy_config);
         Ok(crunchy_config)
     }
 }
@@ -112,11 +120,19 @@ impl Default for IPSConfiguration {
             geolocation_minmax_distance_km: 1000,
             change_at_least: 1,
             change_no_more: 2,
-            location_weight: 0.1,
-            degree_weight: 0.25,
-            eigenvector_weight: 0.25,
-            betweenness_weight: 0.25,
-            closeness_weight: 0.15,
+            mcda_weights: MultiCriteriaAnalysisWeights::default(),
+        }
+    }
+}
+
+impl Default for MultiCriteriaAnalysisWeights {
+    fn default() -> MultiCriteriaAnalysisWeights {
+        MultiCriteriaAnalysisWeights {
+            location: 0.1,
+            degree: 0.25,
+            eigenvector: 0.25,
+            betweenness: 0.25,
+            closeness: 0.15,
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -80,7 +80,6 @@ impl CrunchyConfiguration {
     pub fn new(conf_path: &str) -> Result<CrunchyConfiguration> {
         let config_string = fs::read_to_string(conf_path)?;
         let crunchy_config: CrunchyConfiguration = toml::from_str(&config_string)?;
-        println!("{:?}", crunchy_config);
         Ok(crunchy_config)
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -127,11 +127,11 @@ impl Default for IPSConfiguration {
 impl Default for MultiCriteriaAnalysisWeights {
     fn default() -> MultiCriteriaAnalysisWeights {
         MultiCriteriaAnalysisWeights {
-            location: 0.1,
+            location: 0.3,
             degree: 0.25,
-            eigenvector: 0.25,
+            eigenvector: 0.1,
             betweenness: 0.25,
-            closeness: 0.15,
+            closeness: 0.1,
         }
     }
 }
--- a/src/ips.rs
+++ b/src/ips.rs
@@ -123,13 +123,13 @@ impl Ips {
         // Vector contains IpAddr, node index (from the state.nodes) and rating. We need index just
         // to be able to easily get the node from nodes vector after sorting.
         let mut const_factors = Vec::with_capacity(state.nodes.len());
-        for node_idx in 0..state.nodes.len() {
-            let ip = IpAddr::from_str(state.nodes[node_idx].ip.as_str()).expect(ERR_PARSE_IP);
+        for (idx, node) in state.nodes.iter().enumerate() {
+            let ip = IpAddr::from_str(node.ip.as_str()).expect(ERR_PARSE_IP);
             const_factors.push((
                 ip,
-                node_idx,
+                idx,
                 self.rate_node(
-                    &state.nodes[node_idx],
+                    node,
                     *degrees.get(&ip).expect(ERR_GET_DEGREE), // should be safe to unwrap here as degree hashmap is constructed from the same nodes as the state.nodes
                     *eigenvalues.get(&ip).expect(ERR_GET_EIGENVECTOR), // should be safe to unwrap here as eigenvector hashmap is constructed from the same nodes as the state.nodes
                 ),
--- a/src/ips.rs
+++ b/src/ips.rs
@@ -142,14 +142,13 @@ impl Ips {
             }
 
             // Load peerlist with current connections (we don't want to change everything)
-            for (peer_idx, rating) in peer_ratings.iter().enumerate().take(agraph[node_idx].len()) {
-                let peer = agraph[node_idx][peer_idx];
+            for peer in &agraph[node_idx] {
                 peer_list_entry
                     .list
-                    .push(IpAddr::from_str(state.nodes[peer].ip.as_str()).unwrap());
+                    .push(IpAddr::from_str(state.nodes[*peer].ip.as_str()).unwrap());
 
                 // Remeber current peer ratings
-                curr_peer_ratings.push(*rating);
+                curr_peer_ratings.push(peer_ratings[*peer]);
             }
 
             // 2 - Calculate desired vertex degree
--- a/src/ips.rs
+++ b/src/ips.rs
@@ -47,6 +47,17 @@ pub struct Peer {
     pub list: Vec<IpAddr>,
 }
 
+/// Internal structure for storing peer information
+#[derive(Copy, Clone)]
+struct PeerEntry {
+    /// IP address of the peer
+    pub ip: IpAddr,
+    /// Index of the peer in the state.nodes
+    pub index: usize,
+    /// Ranking of the peer
+    pub rating: f64,
+}
+
 const NORMALIZE_TO_VALUE: f64 = 100.0;
 const NORMALIZE_HALF: f64 = NORMALIZE_TO_VALUE / 2.0;
 const NORMALIZE_2_3: f64 = NORMALIZE_TO_VALUE * 2.0 / 3.0;
@@ -125,15 +136,15 @@ impl Ips {
         let mut const_factors = Vec::with_capacity(state.nodes.len());
         for (idx, node) in state.nodes.iter().enumerate() {
             let ip = IpAddr::from_str(node.ip.as_str()).expect(ERR_PARSE_IP);
-            const_factors.push((
+            const_factors.push(PeerEntry {
                 ip,
-                idx,
-                self.rate_node(
+                index: idx,
+                rating: self.rate_node(
                     node,
                     *degrees.get(&ip).expect(ERR_GET_DEGREE), // should be safe to unwrap here as degree hashmap is constructed from the same nodes as the state.nodes
                     *eigenvalues.get(&ip).expect(ERR_GET_EIGENVECTOR), // should be safe to unwrap here as eigenvector hashmap is constructed from the same nodes as the state.nodes
                 ),
-            ));
+            });
         }
 
         // Iterate over nodes to generate peerlist entry for each node
@@ -143,10 +154,7 @@ impl Ips {
             // Clone const factors for each node to be able to modify them
             let mut peer_ratings = const_factors.clone();
 
-            // Remove node itself to ensure we don't add it to peerlist
-            peer_ratings.retain(|x| x.1 != node_idx);
-
-            let mut curr_peer_ratings: Vec<(IpAddr, usize, f64)> = Vec::new();
+            let mut curr_peer_ratings: Vec<PeerEntry> = Vec::new();
 
             let mut peer_list_entry = Peer {
                 ip: node_ip,
@@ -205,8 +213,11 @@ impl Ips {
                 peers_to_add_count = self.config.change_no_more;
             }
 
+            // Remove node itself to ensure we don't add it to peerlist
+            peer_ratings.retain(|x| x.index != node_idx);
+
             // Sort peers by rating (highest first)
-            curr_peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
+            curr_peer_ratings.sort_by(|a, b| b.rating.partial_cmp(&a.rating).unwrap());
 
             // 4 - Choose peers to delete from peerlist (based on ranking)
             while peers_to_delete_count > 0 && curr_peer_ratings.pop().is_some() {
@@ -216,10 +227,10 @@ impl Ips {
             // 5 - Find peers to add from selected peers (based on rating)
             if peers_to_add_count > 0 {
                 // Sort peers by rating
-                peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
+                peer_ratings.sort_by(|a, b| b.rating.partial_cmp(&a.rating).unwrap());
 
                 // Remove peers that are already in peerlist
-                peer_ratings.retain(|x| !peer_list_entry.list.contains(&x.0));
+                peer_ratings.retain(|x| !peer_list_entry.list.contains(&x.ip));
 
                 let mut candidates = peer_ratings
                     .iter()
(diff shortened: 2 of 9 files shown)


src/ips.rs
```diff
@@ -0,0 +1,445 @@
...
+        };
+    }
+
+    fn rate_node(&self, node: &Node, degree: u32, eigenvalue: f64) -> f64 {
+        // Calculate rating for node
+
+        // Rating is a combination of the following factors:
+        // 1. Degree
+        let mut rating = ((degree as f64 - self.degree_factors.min)
+            / (self.degree_factors.max - self.degree_factors.min)
```
kylegranger, line 388: `struct NormalizationFactors` could have a `lerp` method, that handles the linear interpolation.

src/ips.rs
```diff
@@ -0,0 +1,445 @@
...
+            * self.eigenvector_factors.value)
+            * self.config.eigenvector_weight;
+
+        rating
+    }
+
+    // Very simple algorithm to detect islands.
+    // Take first vertex and do BFS to find all connected vertices. If there are any unvisited vertices
+    // create new island and do BFS one more time. Repeat until all vertices are visited.
+    fn detect_islands(&self, nodes: &[Node], agraph: &AGraph) -> Vec<HashSet<usize>> {
```
kylegranger, line 413: Just something to think about...another todo item.  This should maybe be done in the `compute_betweenness_and_closeness`.  An enumeration of the islands (how many, and each node has an island id (index))

kylegranger (commented):
looks very nice, Piotr.. And I'm going to do a second pass, still trying to grok the whole thing, will reread the pdf.

src/ips.rs
```diff
@@ -0,0 +1,445 @@
...
+            .iter()
+            .min_by(|a, b| a.1.cmp(b.1))
+            .map(|m| m.1)
+            .unwrap_or(&0);
+        let max = *degrees
+            .iter()
+            .max_by(|a, b| a.1.cmp(b.1))
+            .map(|m| m.1)
+            .unwrap_or(&(NORMALIZE_TO_VALUE as u32));
+
```
kylegranger, line 317: For here, and the other similar derivation or min/max property values, we don't want to use the normalization value as defaults: 0 and 100. They only have meaning for normalized values, not for the raw min/max to be found here and similar places.  This should always work, so panic would be better, as sign of a real bug or something else.

src/config.rs
```diff
@@ -41,8 +41,33 @@ pub struct GeoIPConfiguration {
...
+    /// Indicates if configuration should be taken into account
+    pub pref_location: bool,
+    /// True means we should prefer closer peers, false means we should prefer farther peers
+    pub pref_location_closer: bool,
+    /// If pref_location_closer is true, this is the maximum distance in kilometers we should prefer
+    /// closer peers. If pref_location_closer is false, this is the minimum distance in kilometers we
+    /// should prefer farther peers.
+    pub pref_location_distance: u32,
+    /// Indicates how many peers must be changed for each node
+    pub change_at_least: u32,
```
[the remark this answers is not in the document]
asmie: Done!

src/ips.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// List of peers for the node
+    pub peer_list: Vec<IpAddr>,
+}
+
+const NORMALIZE_TO_VALUE: f64 = 100.0;
+#[derive(Default, Clone)]
+struct NormalizationFactors {
+    min: f64,
+    max: f64,
+    value: f64,
```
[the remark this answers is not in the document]
asmie: Done.

src/ips.rs
asmie: Nice catch!
Done :)

src/ips.rs
Rqnsom: We use the `nodes` argument to get its length? Could we remove it, or we could add a TODO comment here to indicate we'll use it for something else later 🕵️‍♀️

And it feels like the function asserts:
```rust
assert_eq!(nodes.len(), agraph.len());
```
Could we add it on top if we need `nodes` for something else later?

src/ips.rs
```diff
@@ -0,0 +1,431 @@
...
+    eigenvector_factors: NormalizationFactors,
+}
+
+/// Peer list structure containing peer list for each node
+#[derive(Clone, Serialize, Deserialize)]
+pub struct PeerList {
+    /// IP address of the node
+    pub peer: IpAddr,
+    /// List of peers for the node
+    pub peer_list: Vec<IpAddr>,
```
Rqnsom, line 40:
```suggestion
pub struct Peer {
    /// IP address of the node
    pub ip: IpAddr,
    /// List of peers for the node
    pub list: Vec<IpAddr>,
```
Or maybe `pub struct PeerInfo {`

Rqnsom (commented):
I just checked the start of the function for now 😅

testdata/config.toml
```diff
@@ -10,4 +10,14 @@ ipapico_enable = false
 ipapicom_enable = false
 
 [ips_config]
-
+peer_file_path = "testdata/peers.json"
+pref_location = true
+pref_location_closer = true
+pref_location_distance = 1000
```
Rqnsom, line 16: Would be nice to have comments in this file, as the real user won't read the code in search of the documentation

testdata/config.toml
```diff
@@ -10,4 +10,14 @@ ipapico_enable = false
...
+change_at_least = 1
+change_no_more = 2
+location_weight = 0.10
+degree_weight = 0.25
+eigenvector_weight = 0.25
+betweenness_weight = 0.25
+closeness_weight = 0.15
```
Rqnsom, line 23: Should we add a check to expect the sum of these is 1.0 in the code?

testdata/config.toml
kylegranger: some fields were renamed (above).  `config.toml` is just not yet updated.
```
use_geolocation: true,
use_closer_geolocation: true,
geolocation_minmax_distance_km: 1000,
```

asmie: Done.

src/ips.rs
asmie: You're right. agraph is sufficent for now (and probably it should be also after moving that function to spectre).
Changed.

src/ips.rs
asmie: Done.

testdata/config.toml
asmie: That could not be truth. 
Eg. some users may want to define weights as 1.0, 1.0, 1.0 and 2.0 to emphasize that some factor is two times important than other and that also be fine. 
Some other users may want to define negative weights to give a penalty for some factor. And that also should be fine.

kylegranger (approved):
LGTM.  Of course, we will iterate on this as we all learn more.

src/ips.rs
```diff
@@ -0,0 +1,523 @@
...
+        let degree_avg = self.degree_centrality_avg(&degrees);
+        let eigenvalues = graph.eigenvalue_centrality();
+
+        // Determine factors used for normalization.
+        // Normalization step is needed to make sure that all the factors are in the same range and
+        // weights can be applied to them.
+        self.determine_degrees_factors(&degrees);
+        self.determine_eigenvalue_factors(&eigenvalues);
+        self.determine_betweenness_factors(&state.nodes);
+        self.determine_closeness_factors(&state.nodes);
```
Rqnsom, line 89: It seems like these four functions are calculating min/max for a given set of numbers (which can have a different type, but all implement [PartialOrd](https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#))

One alternative we could consider could be a generic function with a generic parameter (I'll throw multiple ideas in the example)
```rust
let betweenness = [...]; // collected values from state.node.iter() for betweenness
let closeness = [...]; // collected values from state.node.iter() for closeness

self.degrees_factors = NormalizationFactors::determine(&deegrees.values().cloned().collect());
self.eigenvector_factors = NormalizationFactors::determine(&eigenvalues.values().cloned().collect());
self.betweenness_factors = NormalizationFactors::determine(&betweenness);
self.closeness_factors = NormalizationFactors::determine(&closeness);
```

src/config.rs
```diff
@@ -41,8 +41,33 @@ pub struct GeoIPConfiguration {
...
+    /// Weight (importance) of the location factor (used in multi-criteria analysis)
+    pub location_weight: f64,
+    /// Weight (importance) of the degree factor (used in multi-criteria analysis)
+    pub degree_weight: f64,
+    /// Weight (importance) of the eigenvector factor (used in multi-criteria analysis)
+    pub eigenvector_weight: f64,
+    /// Weight (importance) of the betweenness factor (used in multi-criteria analysis)
+    pub betweenness_weight: f64,
+    /// Weight (importance) of the closeness factor (used in multi-criteria analysis)
+    pub closeness_weight: f64,
```
Rqnsom, line 69: to avoid repeating the same comment `(used in multi-criteria analysis)` 

We could consider to move these values into a mca struct:
```rust
/// Algorithm weight factors.
pub mca: MultiCriteriaAnalysisWeights;
```
now there's no need to add the `weight` keyword to every name
```rust
struct MultiCriteriaAnalysisWeights {  // or just MultiCriteriaWeights
    pub location: f64,
    pub degree_weight: f64,
    [...]
}
```

Rqnsom (commented):
a few more suggestions/ideas

src/ips.rs
```diff
@@ -0,0 +1,431 @@
...
+        for i in 0..agraph.len() {
+            if visited[i] {
+                continue;
+            }
+
+            let mut island = HashSet::new();
+            let mut queue = VecDeque::new();
+            queue.push_back(i);
+
+            while let Some(node_idx) = queue.pop_front() {
```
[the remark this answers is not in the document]
asmie: Unit tests with and without islands added. 
That code search for islands in the meaning of looking if graph is disconnected.
Interconnected islands will be done in next PR (just not to extend this one).

src/ips.rs
```diff
@@ -0,0 +1,523 @@
...
+            }
+        }
+        graph
+    }
+
+    fn degree_centrality_avg(&self, degrees: &HashMap<IpAddr, u32>) -> f64 {
+        let mut sum = 0.0;
+        for degree in degrees.values() {
+            sum += *degree as f64;
+        }
```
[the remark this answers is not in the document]
asmie: Done.

src/ips.rs
```diff
@@ -0,0 +1,523 @@
...
+        self.closeness_factors = NormalizationFactors { min, max };
+    }
+
+    fn rate_node(&self, node: &Node, degree: u32, eigenvalue: f64) -> f64 {
+        // Calculate rating for node
+
+        // Rating is a combination of the following factors:
+        // 1. Degree
+        let mut rating = ((degree as f64 - self.degree_factors.min)
+            / (self.degree_factors.max - self.degree_factors.min)
```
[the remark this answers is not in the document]
asmie: Done (assertion).

src/ips.rs
asmie: With some issues but done :)

src/ips.rs
```diff
@@ -0,0 +1,505 @@
...
+                graph.insert(edge);
+            }
+        }
+        graph
+    }
+
+    fn degree_centrality_avg(&self, degrees: &HashMap<IpAddr, u32>) -> f64 {
+        let sum = degrees
+            .iter()
+            .fold(0.0, |acc, (_, &degree)| acc + degree as f64);
```
Rqnsom, line 327: this fits in one line
```suggestion
        let sum = degrees.iter().fold(0, |acc, (_, &degree)| acc + degree) as f64;
```

src/ips.rs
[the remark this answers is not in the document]
Rqnsom: Sure... but would it be better to skip updating the rating in that case? 

Otherwise, it would be funny to see this in the description:
> How do you know the network reached ideal levels after applying algorithm adjustments? 
> Well, you will know... the algorithm will crash itself if that ever happens

😃

src/config.rs
asmie: Changed. 
As this makes also changes to configuration file _config.toml_ has been also updated.

src/ips.rs
```diff
@@ -0,0 +1,505 @@
...
+        // Determine factors used for normalization.
+        // Normalization step is needed to make sure that all the factors are in the same range and
+        // weights can be applied to them.
+        self.degree_factors = NormalizationFactors::determine(
+            &degrees
+                .values()
+                .cloned()
+                .map(|d| d as f64)
+                .collect::<Vec<f64>>(),
+        );
```
Rqnsom, line 92: we can keep the `u32` - conversion to f64 can happen later since we have Into<64> trait
```suggestion
        self.degree_factors = NormalizationFactors::determine(
            &degrees
                .values()
                .cloned()
                .collect::<Vec<u32>>(),
        );
```
this could fit one line now maybe

src/ips.rs
asmie: Done

src/ips.rs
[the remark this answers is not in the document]
asmie: Omitting factors that could result in zero division.

src/ips.rs
```diff
@@ -0,0 +1,445 @@
...
+                    &state.nodes[node_idx],
+                    *degrees.get(&ip).unwrap(), // should be safe to unwrap here
+                    *eigenvalues.get(&ip).unwrap(),
+                ),
+            ));
+        }
+
+        // Iterate over nodes to generate peerlist entry for each node
+        for (node_idx, node) in state.nodes.iter().enumerate() {
+            let node_ip = IpAddr::from_str(node.ip.as_str()).unwrap();
```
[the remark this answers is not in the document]
asmie: That's a good point - next PR would refactor the main IPS function with new ideas after the meeting. 
In this one I'd not like to introduce more stuff.

src/ips.rs
asmie: Added.

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+            .max_by(|a, b| a.partial_cmp(b).unwrap())
+            .unwrap();
+
+        NormalizationFactors {
+            min: (*min).into(),
+            max: (*max).into(),
+        }
+    }
+
+    fn lerp(&self, value: f64) -> f64 {
```
niklaslong, line 409: Is this linear interpolation? Perhaps some doc would be helpful here as this isn't what is commonly referred to as `lerp` iirc?

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+        // to be able to easily get the node from nodes vector after sorting.
+        let mut const_factors = Vec::with_capacity(state.nodes.len());
+        for node_idx in 0..state.nodes.len() {
+            let ip = IpAddr::from_str(state.nodes[node_idx].ip.as_str()).unwrap();
+            const_factors.push((
+                ip,
+                node_idx,
+                self.rate_node(
+                    &state.nodes[node_idx],
+                    *degrees.get(&ip).unwrap(), // should be safe to unwrap here
```
niklaslong, line 118: > should be safe to unwrap here

Nit: I'd add why and make it obvious with e.g. `Safety: ...`

niklaslong: Perhaps using `expect` with nice error messages is a good way to go, since there as a lot of `unwraps` based on similar assumptions in this code.

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+
+        for (node_idx, node) in nodes.iter().enumerate() {
+            let mut rating = 0.0;
+            if let Some(geo_info) = &node.geolocation {
+                let location = Location::new(
+                    geo_info.latitude.unwrap_or_default(),
+                    geo_info.longitude.unwrap_or_default(),
+                );
+                let distance = selected_location.distance_to(&location).unwrap().meters();
+                let pref_distance = self.config.geolocation_minmax_distance_km as f64 * 1000.0;
```
niklaslong, line 277: As the struct field was changed, perhaps the var should mirror this?

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+                // Remeber current peer ratings
+                curr_peer_ratings.push(peer_ratings[*peer]);
+            }
+
+            // 2 - Calculate desired vertex degree
+            // In the first iteration we will use avg of degree and degree_delta so all
+            // nodes should pursue to degree_delta level. That could be bad if graph's vertexes
+            // have very high (or low) degrees and therefore, delta is very high (or low) too. But until
+            // we have some better idea this one is the best we can do to keep up with the graph.
+            let desired_degree = (degree_avg as u32 + degrees.get(&node_ip).unwrap()) / 2;
```
niklaslong, line 159: Ff the nominator is uneven, there's a chance the division by two could be floored to the closest natural number.

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+
+            // 3 - Calculate how many peers to add or delete from peerlist
+            let peers_to_delete_count = if desired_degree < *degrees.get(&node_ip).unwrap() {
+                degrees.get(&node_ip).unwrap() - desired_degree
```
niklaslong, line 163: Is there any guarantee this won't overflow?

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+                // Check if config forces to change peerlist even if we have good degree.
+                // This should be always set to at least one to allow for some changes in graph -
+                // searching for better potential peers.
+                self.config.change_at_least
+            };
+
+            // Calculating how many peers should be added. If we have more peers than desired degree
+            // we will add at least config.change_at_least peers.
+            let mut peers_to_add_count = if desired_degree > *degrees.get(&node_ip).unwrap() {
+                desired_degree - degrees.get(&node_ip).unwrap() + peers_to_delete_count
```
niklaslong, line 174: ditto

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+
+            // 4 - Choose peers to delete from peerlist (based on ranking)
+            if peers_to_delete_count > 0 {
+                // Sort peers by rating (highest first)
+                curr_peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
+
+                // Remove peers with lowest rating
+                for _ in 0..peers_to_delete_count {
+                    let peer_to_delete = curr_peer_ratings.pop();
+                    if let Some(peer_to_delete) = peer_to_delete {
```
Rqnsom, line 192: if we are deleting, then curr_peer_ratings > peers_to_delete_count so we are safe to use unwrap here directly after `pop()`

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+
+            // Limit number of changes to config value
+            if peers_to_add_count > self.config.change_no_more {
+                peers_to_add_count = self.config.change_no_more;
+            }
+
+            // 4 - Choose peers to delete from peerlist (based on ranking)
+            if peers_to_delete_count > 0 {
+                // Sort peers by rating (highest first)
+                curr_peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
```
Rqnsom, line 187:
```suggestion
        // Sort peers by rating (highest first)
        curr_peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
            // 4 - Choose peers to delete from peerlist (based on ranking)
            while peers_to_delete_count > 0 { // and then we can deduct one at the end of the loop
```

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+                });
+
+                for peer in candidates.iter().take(peers_to_add_count as usize) {
+                    peer_list_entry.list.push(peer.0);
+                }
+            }
+
+            // Do not compute factors one more time after every single peerlist addition. At least
+            // for now, when computing factors is very expensive (especially betweenness and closeness).
+            // Re-calculating it after each node for whole graph would take too long.
```
Rqnsom, line 242: (maybe a TODO idea):
but it would be safe maybe to detect islands for adjusted peer lists to see that the algo hasn't accidentally split the network in half

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+            let mut peers_to_add_count = if desired_degree > *degrees.get(&node_ip).unwrap() {
+                desired_degree - degrees.get(&node_ip).unwrap() + peers_to_delete_count
+            } else {
+                self.config.change_at_least
+            };
+
+            // Limit number of changes to config value
+            if peers_to_add_count > self.config.change_no_more {
+                peers_to_add_count = self.config.change_no_more;
+            }
```
Rqnsom, line 182: maybe - to speed-up the process:
```suggestion
            if peers_to_add_count > self.config.change_no_more {
                peers_to_add_count = self.config.change_no_more;
                peers_to_delete_count = 0;
            }
```

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+                let mut candidates_to_search = peers_to_add_count * 2;
+
+                let mut candidates: Vec<(IpAddr, usize, f64)> =
+                    Vec::with_capacity(candidates_to_search as usize);
+
+                // Sort peers by rating
+                peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
+
+                // Add peers with highest rating as candidates
+                for peer in peer_ratings.iter() {
```
niklaslong, line 211: I think this can perhaps be written more cleanly with iterators? i.e. with `filter` to check for inclusion and then perhaps a `collect`? Just a thought...

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+                    // Check if peer is already in peerlist - if so go to next one
+                    if peer_list_entry.list.contains(&peer.0) {
+                        continue;
+                    }
```
Rqnsom, line 215: Right now, is it ensured that the node cannot add itself to the list? Do we need a check for that?

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+// updated with location factor. Each factor has its own weight that is used to determine
+// factor's importance to the calculation of the final ranking. That gives possibility
+// to test different approaches to the selection of the peers, without re-compiling the code.
+// Weights are defined in the configuration file.
+// When the ranking is calculated, the peers are selected based on the ranking. The number of
+// peers can be changed and it is defined in the configuration file. Algorithm is constructed
+// as step by step process, so it is easy to add new steps or change the order of the steps.
+// Especially, there could be a need to add some modifiers to the ranking.
+
+use crate::config::IPSConfiguration;
```
niklaslong, line 13: Random note on imports: I quite like having them formatted by `rustfmt`, I usually use e.g.: https://github.com/niklaslong/spectre/blob/main/.rustfmt.toml. (note, one of those might require the nightly formatter).

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+            if peers_to_add_count > 0 {
+                // Take twice as many candidates as we need to add to peerlist to be able to
+                // choose best ones from them.
+                let mut candidates_to_search = peers_to_add_count * 2;
+
+                let mut candidates: Vec<(IpAddr, usize, f64)> =
+                    Vec::with_capacity(candidates_to_search as usize);
+
+                // Sort peers by rating
+                peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
```
Rqnsom, line 208: since peer_ratings is copied for every node, we are performing the same sort multiple times

Also, it means we would choose only a handful of top nodes by rating and put those on each suggested peer list where some node 🕵️‍♀️

so if we have 1000 peers, and 500 peers need to add ~5 nodes to improve their average degree,
all 500 peer list suggestions will contain the same (the best) top ~10 (since 5*2) nodes from the total of 1000 nodes

Rqnsom (commented):
final comment, overall LGTM

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    // Helper functions
+
+    /// Update nodes rating based on location
+    fn update_rating_by_location(
+        &self,
+        selected_node: &Node,
+        nodes: &[Node],
+        ratings: &mut [(IpAddr, usize, f64)],
```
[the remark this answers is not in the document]
Rqnsom: I see we sort the `peer_ratings` down there, so the indices won't match anymore. Feel free to resolve this convo when you see it 

Done.

src/ips.rs
asmie: The logic would prevent from overflow (underflow) here but added saturated commands just to prevent it in case if at any point in the future the logic would change.

src/ips.rs
asmie: As above.

src/ips.rs
asmie: Changed.

src/ips.rs
asmie: Added rounding instead of flooring - this could better suit or intention.

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+                if self.config.use_closer_geolocation {
+                    match distance {
+                        _ if distance < pref_distance => rating = NORMALIZE_TO_VALUE,
+                        _ if distance < 2.0 * pref_distance => {
+                            rating = NORMALIZE_TO_VALUE * 2.0 / 3.0
+                        }
+                        _ if distance < 3.0 * pref_distance => {
+                            rating = NORMALIZE_TO_VALUE * 1.0 / 3.0
+                        }
+                        _ => rating = 0.0,
```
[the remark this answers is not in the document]
asmie: It won't be disconnected - no other node would make new connection to it (and if some other nodes need to cut off some connection they probably cut this node out) but this node will make a new connections to other nodes based on the other nodes ranking. So this node will have their own peerlist updated and therefore can connect to other nodes in the network.

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+        // Node rating can be split into two parts: constant and variable depending on the node's
+        // location. Now we can compute each node's constant rating based on some graph params.
+        // Vector contains IpAddr, node index (from the state.nodes) and rating. We need index just
+        // to be able to easily get the node from nodes vector after sorting.
+        let mut const_factors = Vec::with_capacity(state.nodes.len());
+        for node_idx in 0..state.nodes.len() {
+            let ip = IpAddr::from_str(state.nodes[node_idx].ip.as_str()).unwrap();
+            const_factors.push((
+                ip,
+                node_idx,
```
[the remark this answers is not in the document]
asmie: Yes, we are using:
```
candidates.sort_by(|a, b| {
                    state.nodes[a.1]
                        .betweenness
                        .partial_cmp(&state.nodes[b.1].betweenness)
                        .unwrap()
                });
```

Unfortunately, node index is a synchronization index between nodes list, agraph and peer_list. TBH if we'd not need agraph and be able to count everything using adj lists where we can identify each node (as IP) I could try to link peer_ratings in some other way just to nodes list. Agraph for now contains only positions from node's list.

src/ips.rs
asmie: That's not the same sort for each node. 
For each node `peer_ratings` are created from `const_factors` and then, each peer rating is adjusted based on distance between current node and that peer.
That makes `peer_ratings` list unique for each node.

src/ips.rs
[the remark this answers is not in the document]
Rqnsom: aha, yes!

src/ips.rs
asmie: Check added.

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+                let distance = selected_location.distance_to(&location).unwrap().meters();
+                let pref_distance = self.config.geolocation_minmax_distance_km as f64 * 1000.0;
+
+                // Map distance to some levels of rating - now they are taken arbitrarily but
+                // they should be somehow related to the distance.
+                if self.config.use_closer_geolocation {
+                    match distance {
+                        _ if distance < pref_distance => rating = NORMALIZE_TO_VALUE,
+                        _ if distance < 2.0 * pref_distance => {
+                            rating = NORMALIZE_TO_VALUE * 2.0 / 3.0
```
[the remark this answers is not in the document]
asmie: Compiler should handle that but consts added.

src/ips.rs
Rqnsom: right, so my previous comment is true only when the `location` factor is set to zero.

but let me try to think about this like this:
so the `location` factor is `0.1`

which means that the `const_factors` contain a list of ratings from `0 - 90.0` - and that list, once sorted, is pretty static.
then if we calculate location factors for all nodes separately and sort them, we get a list with a rating range: `0 - 10.0`

when we add those two vectors together, the overall node rating won't change that much, since the variable part (the location) only makes a 0-10 difference.

so that means the top 20 best nodes will still be pretty similar in the final `peer_ratings` list.
so I propose we increase the location_ration to a higher value to make sure it breaks the static sorted list of `const_factors` ratings. so the top 20 nodes are much more different from node to node.

src/config.rs
```diff
@@ -40,9 +40,41 @@ pub struct GeoIPConfiguration {
     pub ipapicom_api_key: Option<String>,
 }
 
+/// Multi-criteria analysis weights
+#[derive(Debug, Clone, Deserialize)]
+pub struct MultiCriteriaAnalysisWeights {
+    /// Weight (importance) of the location factor (used in multi-criteria analysis)
+    pub location: f64,
```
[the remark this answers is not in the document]
asmie: Changed.

src/ips.rs
asmie: You're right, changed to a name that would fit better.

src/ips.rs
```diff
@@ -0,0 +1,525 @@
...
+    /// agraph are the same as the positions of the nodes in the state.nodes.
+    pub async fn generate(&mut self, state: &CrunchyState, agraph: &AGraph) -> Vec<Peer> {
+        let mut peer_list = Vec::new();
+
+        // Reconstruct graph from the agraph - we need to do this because we need all the
+        // measures provided by spectre's graph implementation.
+        // Using agraph gives us certainity that we are using the same graph as the crawler and
+        // there are only good nodes there (this is critical assumption!). Second assumption is
+        // that agraph node indexes are the same as in the state.nodes vector.
+        let mut graph = self.construct_graph(&state.nodes, agraph);
```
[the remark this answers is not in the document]
asmie: I think so, currently we're receiving nodes and agraph from crawler and re-constructing edge graph. 
Probably we'd need to modify crawler to eliminate agraph but it's another subject to be done.

src/ips.rs
asmie: Added expect error messages and refactored degree usage a bit later to avoid many unwraps.

src/ips.rs
asmie: Default weights changed.

src/ips.rs
asmie: Yes, definately todo idea as currently we're using agraph to detect islands and agraph is not updated during calculations (agraph may be removed in the near future).

src/ips.rs
asmie: There is possible that we'd like to add peers but there are also some peers to delete - limiting change count does not mean that there are no peers to delete (if there is change_at_least option set we need to delete at least as many peers as there is set).

src/ips.rs
asmie: Done.

src/ips.rs
Rqnsom: Would it be okay if we leave a `TODO` comment in the code as a reminder?

src/ips.rs
asmie: Now code is changed (that part is refactored) and there is possibility to have curr_peer_ratings empty.

src/ips.rs
```diff
@@ -0,0 +1,533 @@
...
+            }
+
+            // Sort peers by rating (highest first)
+            curr_peer_ratings.sort_by(|a, b| b.2.partial_cmp(&a.2).unwrap());
+
+            // 4 - Choose peers to delete from peerlist (based on ranking)
+            while peers_to_delete_count > 0 {
+                // Remove peers with lowest rating
+                if let Some(peer_to_delete) = curr_peer_ratings.pop() {
+                    peer_list_entry.list.retain(|x| x != &peer_to_delete.0);
```
Rqnsom, line 212: `pop()` removes the peer so this should suffice, no?
```suggestion
                if let Some(peer_to_delete) = curr_peer_ratings.pop() {
```

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+                    geo_info.longitude.unwrap_or_default(),
+                );
+                let distance = selected_location.distance_to(&location).unwrap().meters();
+                let pref_distance = self.config.geolocation_minmax_distance_km as f64 * 1000.0;
+
+                // Map distance to some levels of rating - now they are taken arbitrarily but
+                // they should be somehow related to the distance.
+                if self.config.use_closer_geolocation {
+                    match distance {
+                        _ if distance < pref_distance => rating = NORMALIZE_TO_VALUE,
```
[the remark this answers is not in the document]
asmie: Done.

src/ips.rs
asmie: Comment left.

src/config.rs
```diff
@@ -40,9 +40,41 @@ pub struct GeoIPConfiguration {
...
-#[derive(Default, Debug, Clone, Deserialize)]
-pub struct IPSConfiguration {} //TODO(asmie): implement with IPS
+#[derive(Debug, Clone, Deserialize)]
+pub struct IPSConfiguration {
+    /// Path where peer list file will be written
+    pub peer_file_path: Option<PathBuf>,
+    /// Indicates if configuration should be taken into account
+    pub use_geolocation: bool,
+    /// True means we should prefer closer peers, false means we should prefer farther peers
+    pub use_closer_geolocation: bool,
```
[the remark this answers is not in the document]
asmie: Added.

src/ips.rs
asmie: Deletion changed one more time to be simple loop with poping values.

src/ips.rs
```diff
@@ -0,0 +1,526 @@
...
+            }
+
+            // Load peerlist with current connections (we don't want to change everything)
+            for (peer_idx, rating) in peer_ratings.iter().enumerate().take(agraph[node_idx].len()) {
+                let peer = agraph[node_idx][peer_idx];
+                peer_list_entry
+                    .list
+                    .push(IpAddr::from_str(state.nodes[peer].ip.as_str()).unwrap());
+
+                // Remeber current peer ratings
```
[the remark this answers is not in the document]
asmie: Fixed.

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+                Location::new(
+                    geo_info.latitude.unwrap_or_default(), // TODO(asmie): refactor when zg-core will be updated with Location
+                    geo_info.longitude.unwrap_or_default(),
+                )
+            })
+            .unwrap_or(Location::new(0.0, 0.0));
+
+        for (node_idx, node) in nodes.iter().enumerate() {
+            let mut rating = 0.0;
+            if let Some(geo_info) = &node.geolocation {
```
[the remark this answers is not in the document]
asmie: Done.

src/ips.rs
asmie: Formatter added.

src/ips.rs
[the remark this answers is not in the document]
asmie: Changed to struct.

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+                    }
+                } else {
+                    match distance {
+                        _ if distance < 0.5 * pref_distance => rating = 0.0,
+                        _ if distance < pref_distance => rating = NORMALIZE_TO_VALUE / 2.0,
+                        _ => rating = NORMALIZE_TO_VALUE,
+                    }
+                }
+            }
+            ratings[node_idx].2 += rating * self.config.mcda_weights.location;
```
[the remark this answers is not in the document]
asmie: As above :)

src/ips.rs
```diff
@@ -0,0 +1,504 @@
...
+            .map(|n| n.closeness)
+            .collect::<Vec<f64>>();
+        self.closeness_factors = NormalizationFactors::determine(closeness);
+
+        // Node rating can be split into two parts: constant and variable depending on the node's
+        // location. Now we can compute each node's constant rating based on some graph params.
+        // Vector contains IpAddr, node index (from the state.nodes) and rating. We need index just
+        // to be able to easily get the node from nodes vector after sorting.
+        let mut const_factors = Vec::with_capacity(state.nodes.len());
+        for node_idx in 0..state.nodes.len() {
```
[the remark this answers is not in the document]
asmie: Done.

Rqnsom (approved):
LGTM, this is a nice piece of work! 🚀

PR #8, opened 2023-02-20
refactor: use Location struct from newest ziggurat-core

asmie:
Make usage of Location struct introduced in ziggurat-core.

NOTE: this change introduces changes in geo-cache file and output JSON file.

Samples updated respectively under testdata directory.


lwojtow (approved):

kylegranger (approved):
LGTM

Rqnsom (approved):

src/ips.rs
```diff
@@ -273,28 +272,28 @@ impl Ips {
...
-                    geo_info.longitude.unwrap_or_default(),
-                )
-            })
-            .unwrap_or(Location::new(0.0, 0.0));
+        if selected_node.geolocation.is_none() {
+            return;
+        }
+
+        let selected_location =
+            if let Some(location) = selected_node.geolocation.as_ref().unwrap().location {
```
Rqnsom, line 280: but I've been thinking, it feels a bit funny to have something named `location` unwrapped twice.
long term alternative could be:

location
-> city/country name
-> can have other stuff
-> has coordinates

so the `location` it can be a broader term - like our outer struct

while the term coordinates are mainly defined by the longitude and latitude
[Latitude and Longitude are the units that represent the coordinates at geographic coordinate system. ](https://www.latlong.net/)

coordinate
-> longitude
-> latitude

kylegranger: I have to say, I didn't see this before, but I kind of agree. `coordinates` (and `Coordinates`) does seem intuitively better.

asmie: Well, that's no problem to refactor that name.

PR #10, opened 2023-02-21
feat: interconnected islands bridges detection

asmie:
Find bridges in graph.
Bridges are edges that if removed disconnects the graph but here we try to find something similar to bridges - connections that acts like bridges between two inter-connected islandsbut cutting them do not disconnect the graph (as there can be couple of bridges for each interconnected island). That is why we use betweenness centrality to find such connections instead of some popular bridge finding algorithms (like Tarjan's algorithm or chain decomposition).

The idea is to find connections that have high betweenness centrality on both ends. The main problem is meaning of high betweenness centrality. This approach uses median of betweenness centrality of all nodes as a base point for threshold. Then, to eliminate some corner cases (eg. when there are only few nodes with high betweenness centrality and most of the nodes have low factor value what could result in finding too many bridges) we adjust the threshold by const factor read from configuration. There could be different approaches like not using the median but taking value from some percentile (eg. 90th percentile) but this could lead to set threshold to find too many bridges in case of eg. balanced graph (if there are many nodes with similar betweenness centrality taking top 20% would result in finding fake bridges).

This PR introduces bridge detection and ensurance that bridge is never deleted.

### DIFF
--- a/src/graph_utils.rs
+++ b/src/graph_utils.rs
@@ -21,11 +21,17 @@ use crate::{utils::median, Node};
 /// with similar betweenness centrality taking top 20% would result in finding fake bridges).
 pub fn find_bridges(nodes: &[Node], threshold_adjustment: f64) -> HashMap<usize, HashSet<usize>> {
     let mut bridges = HashMap::new();
+
+    // If there are less than 2 nodes there is no point in finding bridges.
+    if nodes.len() < 2 {
+        return bridges;
+    }
+
     let mut betweenness_list = nodes.iter().map(|n| n.betweenness).collect::<Vec<f64>>();
 
     betweenness_list.sort_by(|a, b| a.partial_cmp(b).unwrap());
 
-    let betweenness_median = median(&betweenness_list);
+    let betweenness_median = median(&betweenness_list).unwrap(); // Safe to uwrap as we checked if there are at least 2 nodes.
     let betweenness_threshold = betweenness_median * threshold_adjustment;
 
     for (node_idx, node) in nodes.iter().enumerate() {
--- a/src/graph_utils.rs
+++ b/src/graph_utils.rs
@@ -40,7 +40,7 @@ pub fn find_bridges(nodes: &[Node], threshold_adjustment: f64) -> HashMap<usize,
         }
 
         for peer_idx in &node.connections {
-            if nodes[*peer_idx].betweenness < betweenness_threshold || node_idx == *peer_idx {
+            if nodes[*peer_idx].betweenness <= betweenness_threshold {
                 continue;
             }
 
--- a/src/graph_utils.rs
+++ b/src/graph_utils.rs
@@ -1,0 +1,124 @@
+use std::collections::{HashMap, HashSet};
+
+use crate::{utils::median, Node};
+
+/// Find bridges in graph.
+/// Bridges are edges that if removed disconnects the graph but here we try to find something
+/// similar to bridges - connections that acts like bridges between two inter-connected islands
+/// but cutting them do not disconnect the graph (as there can be couple of bridges for each
+/// interconnected island). That is why we use betweenness centrality to find such connections
+/// instead of some popular bridge finding algorithms (like Tarjan's algorithm or chain
+/// decomposition).
+///
+/// The idea is to find connections that have high betweenness centrality on both ends. The main
+/// problem is meaning of high betweenness centrality. This approach uses median of betweenness
+/// centrality of all nodes as a base point for threshold. Then, to eliminate some corner cases
+/// (eg. when there are only few nodes with high betweenness centrality and most of the nodes have
+/// low factor value what could result in finding too many bridges) we adjust the threshold by
+/// const factor read from configuration. There could be different approaches like not using
+/// the median but taking value from some percentile (eg. 90th percentile) but this could lead to
+/// set threshold to find too many bridges in case of eg. balanced graph (if there are many nodes
+/// with similar betweenness centrality taking top 20% would result in finding fake bridges).
+pub fn find_bridges(nodes: &[Node], threshold_adjustment: f64) -> HashMap<usize, HashSet<usize>> {
+    let mut bridges = HashMap::new();
+    let mut betweenness_list = nodes.iter().map(|n| n.betweenness).collect::<Vec<f64>>();
+
+    betweenness_list.sort_by(|a, b| a.partial_cmp(b).unwrap());
+
+    let betweenness_median = median(&betweenness_list);
+    let betweenness_threshold = betweenness_median * threshold_adjustment;
+
+    for (node_idx, node) in nodes.iter().enumerate() {
+        if node.betweenness < betweenness_threshold {
+            continue;
+        }
+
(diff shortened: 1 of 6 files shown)


src/graph_utils.rs
```diff
@@ -0,0 +1,124 @@
...
+    let betweenness_median = median(&betweenness_list);
+    let betweenness_threshold = betweenness_median * threshold_adjustment;
+
+    for (node_idx, node) in nodes.iter().enumerate() {
+        if node.betweenness < betweenness_threshold {
+            continue;
+        }
+
+        for peer_idx in &node.connections {
+            if nodes[*peer_idx].betweenness < betweenness_threshold || node_idx == *peer_idx {
```
Rqnsom, line 37: possible alternative if this could also work since if they are equal, this check should be true 😃
```suggestion
            if nodes[*peer_idx].betweenness <= betweenness_threshold {
```

src/graph_utils.rs
```diff
@@ -0,0 +1,124 @@
...
+                .or_insert(HashSet::new())
+                .insert(*peer_idx);
+
+            bridges
+                .entry(*peer_idx)
+                .and_modify(|peers: &mut HashSet<usize>| {
+                    peers.insert(node_idx);
+                })
+                .or_insert(HashSet::new())
+                .insert(node_idx);
```
Rqnsom, line 55: this can be inserted later once the algo checks the same connection from the peer point of view (when `peer_idx` is the `node_idx` in the outer loop)
```suggestion
```

src/graph_utils.rs
Rqnsom, line 37: could we just check the `node_idx`'s betweenness with the threshold?

because it's odd to compare relatively to peers here.
in case we have two big islands - and every node in each island is mutually connected with other nodes from that island, and all nodes from both nodes are connected to our bridge node
-> a bridge node would have a huge betweenness
-> everybody else in the network island would have the same betweenness, which could be below the threshold

not sure if we would detect that bridge for that case (where a single node is a bridge between two islands)

src/graph_utils.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+    bridges
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn find_bridges_test() {
```
Rqnsom, line 66: nice! 👍

src/graph_utils.rs
Rqnsom: nvm - now I realize we are storing connections here from node_idx point of view 😅

src/graph_utils.rs
asmie: No, we won't detect them as we can call this node "bridge" node but it's not a graph bridge. 
Bridge is not a node but an edge so it demands to have two nodes on its end. Like any other bridge finding algorithm (Tarjan's or other) it focuses on finding edges that once we cut, we decrease some graph's parameters (classically create disconnected graph).

The problem with one node and many connections to that node from both island's sides does not fit into bridge problem as cutting any of the connections won't make graphs much worse as still we have plenty of other connections to that node.

The problem would be removing that single node.

IMHO it's different problem to solve as that node is more like "hot" node in the network but it's not a bridge itself.

asmie: And just to be precise - hot node problem should be addressed and solved as well, but as it's different from what I was doing here, it should be other feature.

src/utils.rs
```diff
@@ -0,0 +1,15 @@
+/// Calculates the median of a list of numbers.
+pub fn median<T>(list: &[T]) -> f64
+where
```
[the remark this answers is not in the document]
asmie: Nice catch! 
Fixed :)

src/graph_utils.rs
Rqnsom: > IMHO it's different problem to solve...

okay, I see your point and can agree here, but:

> as that node is more like "hot" node in the network, but it's not a bridge itself.

no matter what we call this node, it's an equally important issue we should address here somewhere (it fits into the same category as the general problem trying to be solved), it definitely represents a sort of bridge between two islands (not sure how else we can call it - but if we kill it, it splits the network in two, no?)

the main point is to make sure there is no way to fragment the network into parts - and for that reason, we should detect the interconnected islands and break them.

now we can define that in order to detect interconnected islands, we have two possible scenarios:
   - via connection-bridges (_graph edges_) - so the critical connections which importance needs to be addressed (to make them less important, which is the goal)
   - via node-bridges (_graph nodes/points_?) - so the critical nodes in graph... (same as above)

now, how do we fit this into the whole algorithm?
should we write a `detect_interconnected_islands` function and have two separate functions to handle these two scenarios separately? what do you think the next step should be?

> This PR introduces bridge detection and ensurance that bridge is never deleted.

Sorry for the long comment. I haven't seen this in the IPS.pdf, nor was it mentioned in the PR as a TODO item

asmie: That's right.
Done.

kylegranger: I'm trying to imagine the betweenness values for all the nodes, sorted like a histogram -- e.g., like when you view the black-white values of an image, showing you relatively where the various greys are, with some kind of gridding applied, e.g., 100 values linearly divided between min and max.
It seems to me, the bridge nodes (which I am guessing would be a pair, with very similar values), would have a large delta compared to the **median** betweenness value, e.g., 2.0-3.0x.  A key indicator here (I think) would be a large gap in the histogram, between these two at the top, and the rest of the nodes.  
If the histogram is gradual (i.e., a gaussian), then there doesn't seem to be a smoking gun, showing the existence of a bridge edge (and pair of nodes).

asmie: @kylegranger I thought about that solution. The problem I've noticed in some cases where there will be more "steps" (large deltas) between median and max value. Should we consider first step as a threshold? Or the last one?

Moreover, I don't think if nodes on the both ends of the bridge would have similar values of betweenness. When we've three or four interconnected islands I can image a situation where  one node can have much more betweenness than the other node (because eg. node X connectes islands 1 to island 2 and 3, and node Y connectes only island 1 and 2 - node X would have much more paths because it connectes one island to two another islands and the second one only to one island), but both nodes should have betweenness above the median.

@Rqnsom I'll address your comment shortly :)

Rqnsom (approved):
LGTM 👍

I think we can merge the `find_bridges` function, and then any algo modifications regarding the island detection can be done later separately - or you decide

src/graph_utils.rs
Rqnsom: Nice discussion tday. Resolving this discussion here

src/utils.rs
```diff
@@ -0,0 +1,45 @@
...
+    if list.is_empty() {
+        return None;
+    }
+
+    let mut list = list.to_vec();
+    list.sort_by(|a, b| a.partial_cmp(b).unwrap());
+
+    let mid = list.len() / 2;
+    if list.len() % 2 == 0 {
+        Some((list[mid - 1].into() + list[mid].into()) / 2.0)
```
kylegranger, line 15: 👍

kylegranger (approved):
very nice

PR #11, opened 2023-02-24
refactor: remove agraph usage

asmie:
Remove agraph dependency - agraphs has been removed from spectre and zg-core is also not using it anymore. 
Moved re-generating graph function to graph utils (with agraph removal).
Updated tests (beside one test in main).


src/graph_utils.rs
```diff
@@ -64,10 +70,61 @@ pub fn find_bridges(nodes: &[Node], threshold_adjustment: f64) -> HashMap<usize,
     bridges
 }
 
+pub fn construct_graph(nodes: &[Node]) -> Graph<IpAddr> {
+    let mut graph = Graph::new();
+
+    for i in 0..nodes.len() {
+        for j in 0..nodes[i].connections.len() {
+            let edge = Edge::new(
+                IpAddr::from_str(nodes[i].ip.as_str()).expect(ERR_PARSE_IP),
```
Rqnsom, line 79: one improvement could be to create this IpAddr between lines 76 and 77 since this is a static value between all j iterations and then only copy it here

src/ips.rs
```diff
@@ -85,21 +81,17 @@ impl Ips {
...
         let mut peer_list = Vec::new();
 
-        // Reconstruct graph from the agraph - we need to do this because we need all the
-        // measures provided by spectre's graph implementation.
-        // Using agraph gives us certainity that we are using the same graph as the crawler and
-        // there are only good nodes there (this is critical assumption!). Second assumption is
-        // that agraph node indexes are the same as in the state.nodes vector.
-        let mut graph = self.construct_graph(&state.nodes, agraph);
+        // Reconstruct graph from the state.nodes vector
+        let mut graph = construct_graph(&state.nodes);
```
Rqnsom, line 88: nice!

src/nodes.rs
```diff
@@ -73,30 +73,41 @@ pub fn get_cell_position(
...
 ) -> Vec<Node> {
-    let graph: Graph<usize> = Graph::new();
-    // TODO(asmie/kylegranger): make this an associated function for Graph.
-    // it does not use the Graph per se
-    let (betweenness, closeness) = graph.compute_betweenness_and_closeness(agraph);
-    let mut nodes = Vec::with_capacity(agraph.len());
+    let mut graph = Graph::new();
+    for (n, node) in indices.iter().enumerate() {
+        for connection in node {
+            if *connection > n {
```
Rqnsom, line 83: a mild suggestion if you want to have fun with Rust:
maybe it could be done in an idiomatic way with `node.iter().filter().for_each()` or something (or not sure which method-combination would work) 😃

src/graph_utils.rs
```diff
@@ -64,10 +70,61 @@ pub fn find_bridges(nodes: &[Node], threshold_adjustment: f64) -> HashMap<usize,
     bridges
 }
 
+pub fn construct_graph(nodes: &[Node]) -> Graph<IpAddr> {
+    let mut graph = Graph::new();
+
+    for i in 0..nodes.len() {
```
Rqnsom, line 76: `iter().enumarate()` could be a more Rusty solution maybe

Rqnsom (approved):
LGTM 👍

src/graph_utils.rs
asmie: That's true 👍 Fixed.

src/ips.rs
asmie: 👍

src/graph_utils.rs
asmie: Done.

src/nodes.rs
asmie: Changed.
@kylegranger when you'll be back please look at that change in your code - I hope I've expressed the same what was your intention.

Future remark @kylegranger : I've made a `construct_graph` function to be used inside IPS to create graph with IP addrs. Your loop here does similar thing but constructing graph with node indexes. Maybe we'd create common solution for graph reconstruction or use one graph with some struct inside (having IP and index in it).

PR #12, opened 2023-02-27
debug: visualizer for ips-peers-sample-output.json file

asmie:
Debug util to visualize peers from peerlist. Can be used to visualize previous peerlist and current peerlist just to compare how graph has changed.

Number of connections shown can be limited to speed up rendering process as well as other graph parameters can be changed (image size etc).


kylegranger (approved):
👍

testdata/visualize.py
```diff
@@ -0,0 +1,66 @@
...
+
+for node in data:
+    idx = graph.add_node(node['ip'])
+    ipdict[node['ip']] = idx
+
+conns = 0
+for node in data:
+    node_id = ipdict[node['ip']]
+    for peer_ip in node['list']:
+        graph.add_edge(node_id, ipdict[peer_ip], None)
```
Rqnsom, line 34: should the `conns` be set to zero before we start checking peers here for each node? 🕵️‍♀️

Rqnsom (approved):
maybe it would be nice to have an example image in the repo or something? or on Slack?

testdata/visualize.py
asmie: Fixed. For small number of connections selected it has not generated bad results that's why missed it.

asmie:
Example image added in `testdata` directory.

testdata/visualize.py
Rqnsom: > Fixed

Actually, can you check if you pushed that change, maybe `git add` didn't include that file

asmie: Should be pushed right now.

PR #14, opened 2023-02-27
refactor: cutting main functions (write_state, generate) into smaller ones (1/2)

asmie:
This is the first refactoring commit (out of two) which cut peerlist generation into smaller pieces which made possible to split generation into two main parts:

- security one - where will be put: finding islands, bridges, hot nodes detection, checking if removing some nodes can fragment network;
- optimization one - MCDA.

### DIFF
--- a/src/ips.rs
+++ b/src/ips.rs
@@ -394,7 +394,7 @@ impl Ips {
     }
 
     fn degree_centrality_avg(&self, degrees: &HashMap<IpAddr, u32>) -> f64 {
-        if degrees.len() < 1 {
+        if degrees.is_empty() {
             return 0.0;
         }
 
@@ -531,10 +531,14 @@ mod tests {
             },
         ];
 
-        let peer = ips.generate_peerlist_for_node(&nodes.get(0).unwrap(), &nodes);
+        let peer = ips.generate_peerlist_for_node(nodes.get(0).unwrap(), &nodes);
         assert_eq!(peer.list.len(), 2);
-        assert!(peer.list.contains(&IpAddr::from_str(&nodes.get(1).unwrap().ip).unwrap()));
-        assert!(peer.list.contains(&IpAddr::from_str(&nodes.get(2).unwrap().ip).unwrap()));
+        assert!(peer
+            .list
+            .contains(&IpAddr::from_str(&nodes.get(1).unwrap().ip).unwrap()));
+        assert!(peer
+            .list
+            .contains(&IpAddr::from_str(&nodes.get(2).unwrap().ip).unwrap()));
     }
 
     #[test]
@@ -562,7 +566,7 @@ mod tests {
 
         let state = ips.generate_state(&nodes);
 
-        assert_eq!(ips.rate_node(&nodes.get(0).unwrap(), &state), 10.0);
+        assert_eq!(ips.rate_node(nodes.get(0).unwrap(), &state), 10.0);
     }
 
     #[test]
--- a/src/ips.rs
+++ b/src/ips.rs
@@ -28,10 +28,19 @@ use crate::{
 #[derive(Default, Clone)]
 pub struct Ips {
     config: IPSConfiguration,
-    degree_factors: NormalizationFactors,
-    betweenness_factors: NormalizationFactors,
-    closeness_factors: NormalizationFactors,
-    eigenvector_factors: NormalizationFactors,
+}
+
+/// State structure containing all the information about the graph and nodes at some point
+#[derive(Default, Clone)]
+struct IpsState {
+    pub nodes: Vec<Node>,
+    pub peer_list: Vec<Peer>,
+    pub degrees: HashMap<IpAddr, u32>,
+    pub eigenvalues: HashMap<IpAddr, f64>,
+    pub degree_factors: NormalizationFactors,
+    pub betweenness_factors: NormalizationFactors,
+    pub closeness_factors: NormalizationFactors,
+    pub eigenvector_factors: NormalizationFactors,
 }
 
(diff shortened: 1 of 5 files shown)


src/ips.rs
```diff
@@ -63,87 +72,60 @@ pub const ERR_PARSE_IP: &str = "failed to parse IP address";
...
-        Ips {
-            config,
-            ..Default::default()
-        }
+        Ips { config }
     }
 
     /// Generate peer list - main function with The Algorithm
     /// It needs state and agraph to be passed as parameters which need to be correlated with
     /// the crawler's state and agraph (and with each other), so the indexes saved in the
```
kylegranger, line 88: I think we should phase out use of the word `agraph`, since it's now gone from `spectre`.  I would suggest using `indices` or `indices struct` instead, which refers to `Vec<Vec<usize>>`.   
Also, in the future, maybe we should make `Indices` or `NodeIndices` an explicit type for this structure, defined in `ziggurat-core/ziggurat-core-crawler`?  [outside the scope of this PR.]

src/ips.rs
```diff
@@ -251,28 +236,110 @@ impl Ips {
...
+
+        let mut graph = construct_graph(nodes);
+        let betweenness = graph.betweenness_centrality();
+        let closeness = graph.closeness_centrality();
+
+        // Recalculate factors with new graph
+        for node in ips_state.nodes.iter_mut() {
+            let ip = IpAddr::from_str(&node.ip).unwrap();
+            node.betweenness = *betweenness.get(&ip).unwrap_or(&0.0);
+            node.closeness = *closeness.get(&ip).unwrap_or(&0.0);
```
kylegranger, line 301: If betweenness or closeness cannot be found for an IP address, wouldn't that be a bug, or a sign of some other pathological condition?  I'm not sure we should ignore that behaviour.
Or rather, can this happen if metrics have not yet been generated for a given node?

src/ips.rs
asmie: Removed agraph from comment. 

It could be a good way to move `Indices` to the zg-core just to ensure all apps are using the same structures.

src/ips.rs
asmie: Well, we're generating those factors for each node in graph just a couple of lines above so I cannot find any reason, why some nodes would not have those factors calculated. Therefore, you're right we can treat it like bug.

Changed to `expect`.

kylegranger (approved):
LGTM

src/ips.rs
Rqnsom: > It could be a good way to move Indices to the zg-core just to ensure all apps are using the same structures.

we could move such struct it to spectre, since those term originate from that repo, e.g.:
```rust
    pub fn get_adjacency_indices(&mut self) -> Vec<Vec<usize>> {
```

`usize` ===> `Index`
`Vec<Vec<usize>> `  -> `Vec<Vec<Index>>` ===> `Indices` (maybe)

then we can reference that struct definition from our repos

src/ips.rs
```diff
@@ -251,29 +223,84 @@ impl Ips {
...
+                .collect::<Vec<f64>>(),
+        );
+
+        let betweenness = &nodes.iter().map(|n| n.betweenness).collect::<Vec<f64>>();
+        ips_state.betweenness_factors = NormalizationFactors::determine(betweenness);
+
+        let closeness = &nodes.iter().map(|n| n.closeness).collect::<Vec<f64>>();
+        ips_state.closeness_factors = NormalizationFactors::determine(closeness);
+
+        ips_state.peer_list = Peer::generate_peerlist(nodes);
```
Rqnsom, line 285: I was not sure if this is generating only one peer list or not (from the function signature)
Maybe:
```suggestion
        ips_state.peer_list = Peer::generate_peerlists(nodes);
```
or maybe `generate_all_peerlists`

src/ips.rs
```diff
@@ -251,29 +223,84 @@ impl Ips {
...
+
+        ips_state
+    }
+
+    /// Calculates const factors for each node.
+    fn calculate_const_factors(&self, state: &IpsState) -> Vec<PeerEntry> {
+        let mut const_factors = Vec::with_capacity(state.nodes.len());
+        for (idx, node) in state.nodes.iter().enumerate() {
```
Rqnsom, line 293:
```suggestion
        let mut const_factors = Vec::with_capacity(state.nodes.len());
        
        for (idx, node) in state.nodes.iter().enumerate() {
```

Rqnsom (approved):
Nice, LGTM!

src/ips.rs
asmie: Done.

src/ips.rs
```diff
@@ -251,29 +223,85 @@ impl Ips {
...
+                .collect::<Vec<f64>>(),
+        );
+
+        let betweenness = &nodes.iter().map(|n| n.betweenness).collect::<Vec<f64>>();
+        ips_state.betweenness_factors = NormalizationFactors::determine(betweenness);
+
+        let closeness = &nodes.iter().map(|n| n.closeness).collect::<Vec<f64>>();
+        ips_state.closeness_factors = NormalizationFactors::determine(closeness);
+
+        ips_state.peer_list = Peer::generate_all_peerlists(nodes);
```
Rqnsom, line 285:
```suggestion
        ips_state.peer_lists = Peer::generate_all_peerlists(nodes);
```

Rqnsom: aha, right! it makes sense

src/ips.rs
asmie: As agreed on Slack, we'd move it to ziggurat-core.

PR #16, opened 2023-02-28
refactor: put ips into separate module

asmie:
IPS has been separated from the rest of crunchy.
This move has two major benefits:
- separated code is more readable and don't mix with other crunchy functionalities;
- allow to grab IPS as a separate product more easily.


src/main.rs
```diff
@@ -16,7 +12,7 @@ use ziggurat_core_crawler::summary::NetworkSummary;
 use crate::{
     config::CrunchyConfiguration,
     geoip_cache::GeoIPCache,
-    ips::Ips,
+    ips::ips_algorithm::Ips,
```
Rqnsom, line 15: one less keyword, wdyt?
```suggestion
    ips::algorithm::Ips,
```

Rqnsom (approved):
LGTM

src/main.rs
asmie: Done.

kylegranger (approved):
👍

PR #17, opened 2023-03-01
refactor: remove current peerlist

asmie:
That would allow to compare algorithm state at any point by just re-calulcating state.
To achieve this we simply ensure working only on node's connections instead of separate peer lists.


kylegranger (approved):
👍

PR #20, opened 2023-03-03
fix: when creating new connection add it to the other's side list also

asmie:
Adding new connection cannot be done only on one side but also on the peer side. This allows to trace how many connections has been already added, preventing nodes to add too many connections to highest rated nodes. Lack of this fix resulted in creating couple hot nodes with very high degrees, because their rating was high enough for about 40% nearest nodes. Moreover, lack of this fix resulted in omitting hard limit of one node's changes.

Added also starting check for node's connections consistency checking:
- if there are any nodes connected to themselves;
- if there are any nodes connected to some peers that haven't that node in their connection list.

### DIFF
--- a/src/ips/algorithm.rs
+++ b/src/ips/algorithm.rs
@@ -78,10 +78,7 @@ impl Ips {
         // have the node in their connections.
         for (idx, node) in state.nodes.iter().enumerate() {
             if node.connections.contains(&idx) {
-                println!(
-                    "{} is connected to itself.",
-                    node.addr
-                );
+                println!("{} is connected to itself.", node.addr);
             }
 
             for peer in &node.connections {
@@ -196,9 +193,7 @@ impl Ips {
             // Remove potential peers identified to have too high degree and have already
             // been processed by the algorithm
             peer_ratings.retain(|x| {
-                final_state.nodes[x.index]
-                    .connections
-                    .len()
+                final_state.nodes[x.index].connections.len()
                     < working_state.nodes[x.index].connections.len()
             });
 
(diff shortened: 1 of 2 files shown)


src/ips/algorithm.rs
```diff
@@ -74,6 +74,23 @@ impl Ips {
 
     /// Generate peer list - main function with The Algorithm
     pub async fn generate(&mut self, state: &CrunchyState) -> Vec<Peer> {
+        // Sanity check that each node is really connected to it's peers and the peers also
```
lwojtow, line 77:
```suggestion
        // Sanity check that each node is really connected to its peers and the peers also
```

asmie: True 👍 Fixed.

kylegranger: I'm thinking about this point:
> Added also starting check for node's connections consistency checking:
if there are any nodes connected to themselves;
if there are any nodes connected to some peers that haven't that node in their connection list.

I'm wondering if that filtering should be done during the creation of the output CrunchyState.  I.e., that it should be assumed that CrunchyState is self-consistent and well-formed?

kylegranger: seems to be orthogonal to IPS

asmie: Yes, at some point it'd be good to have consistent CrunchyState (IPS would not deal with that). That can be additional step in crunchy main, before calling IPS to filter out such self-connections or at least warn about potential asymmetrical ones.

kylegranger (approved):
LGTM

src/ips/algorithm.rs
```diff
@@ -74,6 +74,23 @@ impl Ips {
...
+
+            for peer in &node.connections {
+                if !state.nodes[*peer].connections.contains(&idx) {
+                    println!(
+                        "{} is not connected to {} but {} have a connection to it",
+                        node.addr, state.nodes[*peer].addr, node.addr
+                    );
+                }
+            }
+        }
```
Rqnsom, line 92: we only print these cases? nice!

asmie: Yes, as talked on Slack thread, IPS should receive filtered input from such things. So, unless decision will change it's enough to just print information about such cases on input.

IPS prevents such situations on its output by:
- removing node itself from potential peer vector before making any selection;
- implementing this PR which fixed issue when only one side of the connection has been created 🥲  ;
- deleting duplicates from the final peerlist for each node.

PR #21, opened 2023-03-06
feat: added check if network can be easily fragmented by removing hot nodes

asmie:
Network fragmentation cases.

Checking if network can be easily fragmented by attacking 10% of the nodes and if so, preventing such case by creating new connections between their neighbours.

Checking for massive islands in the input data. Merging two massive islands can be risky and shouldn't be done automatic. Both islands could be disconnected for a long time and therefore they can have their own blockchain history and their own reality.


kylegranger (approved):
LGTM 👍

PR #22, opened 2023-03-07
Export histogram data for visualizer.

kylegranger:
This new functionality analyzes the centrality data (betweenness, closeness, and degree), creating distribution tables for the renderer.

### DIFF
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -33,20 +33,24 @@ impl Histogram {
                 self.max = *value;
             }
         }
+
         let delta = self.max - self.min;
         let mut counts = vec![0; num_counts];
         let mut max_count: usize = 0;
+
         if delta == 0.0 {
             return (counts, max_count);
         }
+
         for value in self.values.iter() {
             let part = (*value - self.min) / delta;
-            let mut slot = (part * 256.0).floor() as usize;
+            let mut slot = (part * num_counts as f64).floor() as usize;
             if slot >= num_counts {
                 slot = num_counts - 1;
             }
             counts[slot] += 1;
         }
+
         for count in counts.iter() {
             if *count > max_count {
                 max_count = *count;
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -130,5 +130,4 @@ mod tests {
         assert_eq!(counts.len(), 0);
         assert_eq!(max_count, 0);
     }
-
 }
--- a/src/histogram.rs
+++ b/src/histogram.rs
@@ -41,7 +41,7 @@ impl Histogram {
         let mut counts = vec![0; num_counts];
         let mut max_count: usize = 0;
 
-        if delta == 0.0 {
+        if delta == 0.0 || num_counts == 0 {
             return (counts, max_count);
         }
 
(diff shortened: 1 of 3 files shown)


src/nodes.rs
```diff
@@ -5,7 +5,19 @@ use spectre::{edge::Edge, graph::Graph};
...
 use ziggurat_core_geoip::geoip::GeoInfo;
 
-use crate::geoip_cache::GeoIPCache;
+use crate::{geoip_cache::GeoIPCache, histogram::Histogram};
+
+const HISTOGRAM_COUNTS: usize = 256;
+
+#[derive(Default, Clone, Serialize, Deserialize)]
+pub struct HistogramSummary {
+    /// Minimum value of a factor.
```
asmie, line 14: Is this comment correct for label?

kylegranger: whoops!  now fixed.

src/histogram.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    // Store a single value
+    pub fn add(&mut self, value: f64) {
+        self.values.push(value);
+    }
+
+    // Compute the histogram tables, returning that table along
+    // with the value of the maximum count in a single slot.
+    pub fn compute(&mut self, num_counts: usize) -> (Vec<usize>, usize) {
+        for value in self.values.iter() {
```
asmie, line 31: Nice 👍 . MinMax in single iteration - probably quicker than calling:
`self.values.iter().max()` and `self.values.iter().min()`

src/histogram.rs
```diff
@@ -0,0 +1,118 @@
...
+            return (counts, max_count);
+        }
+
+        for value in self.values.iter() {
+            let part = (*value - self.min) / delta;
+            let mut slot = (part * num_counts as f64).floor() as usize;
+            if slot >= num_counts {
+                slot = num_counts - 1;
+            }
+            counts[slot] += 1;
```
asmie, line 54: Probably never happen as we'd not have such long histograms, just to know that there could be construction like:
`counts[slot].saturating_add(1);`
to prevent overflow and counting from zero 😃

src/histogram.rs
```diff
@@ -0,0 +1,118 @@
...
+
+        if delta == 0.0 {
+            return (counts, max_count);
+        }
+
+        for value in self.values.iter() {
+            let part = (*value - self.min) / delta;
+            let mut slot = (part * num_counts as f64).floor() as usize;
+            if slot >= num_counts {
+                slot = num_counts - 1;
```
asmie, line 52: Could you try running such test case?

```
    #[test]
    fn histogram_zero_test() {
        let mut histogram = Histogram {
            ..Histogram::default()
        };

        histogram.add(1.6);
        histogram.add(2.6);
        histogram.add(3.6);
        histogram.add(4.6);
        let (counts, max_count) = histogram.compute(0);
    }
```

asmie: `saturating_sub()` should be here to prevent `panicked at 'attempt to subtract with overflow'`

asmie (approved):
👍 with one case to be fixed.

src/histogram.rs
kylegranger: Your test is added (🙏 ) (and amended), and passes.  The end looks like this:
```
        histogram.add(1.6);
        histogram.add(2.6);
        histogram.add(3.6);
        histogram.add(4.6);
        let (counts, max_count) = histogram.compute(0);
        assert_eq!(counts.len(), 0);
        assert_eq!(max_count, 0);
```

But I don't understand your comments about using `saturating_sub()`.   On what line did you mean?

kylegranger: Ah, now I understand (duh).  In any event, if `num_counts` is 0, I exit early.

asmie: Exactly :) Exiting early was one of the solutions 👍

PR #24, opened 2023-03-09
feat: add logging to the file

asmie:
Everything done by IPS can now be logged to file (if file is specified in the configuration file) or displayed on screen. When changing println! to writeln! added percentages in statistics delta.

Sample log file added to the testdata directory.

### DIFF
--- a/src/ips/algorithm.rs
+++ b/src/ips/algorithm.rs
@@ -86,29 +86,26 @@ impl Ips {
             None => Ok(Box::new(io::stdout()) as Box<dyn Write>),
         };
 
-        let mut output = match output {
-            Ok(output) => output,
-            Err(e) => {
-                println!("Failed to open log file: {}", e);
-                Box::new(io::stdout()) as Box<dyn Write>
-            }
-        };
+        let mut o = output.unwrap_or_else(|e| {
+            println!("Failed to open the log file: {e}");
+            Box::new(io::stdout()) as Box<dyn Write>
+        });
 
         // Sanity check that each node is really connected to its peers and the peers also
         // have the node in their connections.
-        writeln!(output, "IPS algorithm started... ").unwrap();
+        writeln!(o, "IPS algorithm started...").unwrap();
         let start_time = std::time::Instant::now();
 
-        writeln!(output, "Checking for nodes connected to themselves...").unwrap();
+        writeln!(o, "Checking for nodes connected to themselves...").unwrap();
         for (idx, node) in state.nodes.iter().enumerate() {
             if node.connections.contains(&idx) {
-                writeln!(output, "{} is connected to itself.", node.addr).unwrap();
+                writeln!(o, "{} is connected to itself.", node.addr).unwrap();
             }
 
             for peer in &node.connections {
                 if !state.nodes[*peer].connections.contains(&idx) {
                     writeln!(
-                        output,
+                        o,
                         "{} is not connected to {} but {} have a connection to it",
                         node.addr, state.nodes[*peer].addr, node.addr
                     )
@@ -117,11 +114,7 @@ impl Ips {
             }
         }
 
-        writeln!(
-            output,
-            "Generating initial network state and its statistics... "
-        )
-        .unwrap();
+        writeln!(o, "Generating initial network state and its statistics... ").unwrap();
 
         // This is the working set of factors.
         let mut working_state = self.generate_state(&state.nodes, false);
@@ -129,11 +122,11 @@ impl Ips {
 
         let initial_statistics = generate_statistics(&working_state);
 
-        writeln!(output, "Statistics for the initial network:").unwrap();
-        print_statistics(&initial_statistics, &mut output);
+        writeln!(o, "Statistics for the initial network:").unwrap();
+        print_statistics(&mut o, &initial_statistics);
 
         writeln!(
-            output,
+            o,
             "Generated initial state and statistics in {} s",
             start_time.elapsed().as_secs()
         )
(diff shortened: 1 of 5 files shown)


testdata/sample_ips.log
```diff
@@ -0,0 +1,337 @@
...
+Comparing if network parameters got changed on plus or minus:
+Deltas for given statistics pair:
+----------------------------------------
+Nodes count: 0 (0.000%)
+
+Degree measures:
+Average: 1.5627022653721951 (0.449%)
+Median: 2 (0.552%)
+Min: 2 (50.000%), max: -4 (-0.855%), delta: -6 (-1.293%)
+
```
Rqnsom, line 320: a plus sign for increasing delta would be also nice 
```suggestion
Min: +2 (+50.000%), max: -4 (-0.855%), delta: -6 (-1.293%)
```

src/ips/algorithm.rs
```diff
@@ -77,30 +80,64 @@ impl Ips {
...
+        // Set up logging
+        let output = match self.config.log_path {
+            Some(ref path) => File::create(path).map(|f| Box::new(f) as Box<dyn Write>),
+            None => Ok(Box::new(io::stdout()) as Box<dyn Write>),
+        };
+
+        let mut output = match output {
+            Ok(output) => output,
+            Err(e) => {
+                println!("Failed to open log file: {}", e);
```
Rqnsom, line 92:
```suggestion
                println!("Failed to open the log file: {e}");
```

src/ips/algorithm.rs
```diff
@@ -77,30 +80,64 @@ impl Ips {
...
     /// Generate peer list - main function with The Algorithm
     pub async fn generate(&mut self, state: &CrunchyState) -> Vec<Peer> {
+        // Set up logging
+        let output = match self.config.log_path {
+            Some(ref path) => File::create(path).map(|f| Box::new(f) as Box<dyn Write>),
+            None => Ok(Box::new(io::stdout()) as Box<dyn Write>),
+        };
+
+        let mut output = match output {
+            Ok(output) => output,
```
Rqnsom, line 90: If interested, see if we can replace this match with the [unwrap_or_else](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap_or_else) fn

src/ips/algorithm.rs
```diff
@@ -77,30 +80,64 @@ impl Ips {
...
+            Err(e) => {
+                println!("Failed to open log file: {}", e);
+                Box::new(io::stdout()) as Box<dyn Write>
+            }
+        };
+
         // Sanity check that each node is really connected to its peers and the peers also
         // have the node in their connections.
+        writeln!(output, "IPS algorithm started... ").unwrap();
```
Rqnsom, line 99:
```suggestion
        writeln!(output, "IPS algorithm started...").unwrap();
```

src/ips/algorithm.rs
```diff
@@ -77,30 +80,64 @@ impl Ips {
...
+        let start_time = std::time::Instant::now();
+
+        writeln!(output, "Checking for nodes connected to themselves...").unwrap();
         for (idx, node) in state.nodes.iter().enumerate() {
             if node.connections.contains(&idx) {
-                println!("{} is connected to itself.", node.addr);
+                writeln!(output, "{} is connected to itself.", node.addr).unwrap();
```
Rqnsom, line 105: `f` just because we repeat this call too many times, here's a suggestion to rename output to `f` as in `file`
or `o` as in `output`

one can easily check what it means at the function start - but usually one assumes it's either a file or 
 the stdout
```suggestion
                writeln!(f, "{} is connected to itself.", node.addr).unwrap();
```

src/ips/algorithm.rs
```diff
@@ -77,30 +80,64 @@ impl Ips {
...
         // This is the working set of factors.
         let mut working_state = self.generate_state(&state.nodes, false);
         let mut final_state = working_state.clone();
 
         let initial_statistics = generate_statistics(&working_state);
-        println!("Statistics for the initial network:");
-        print_statistics(&initial_statistics);
+
+        writeln!(output, "Statistics for the initial network:").unwrap();
+        print_statistics(&initial_statistics, &mut output);
```
Rqnsom, line 133: to reflect the above `std::fs` APIs, the `fs`/`ouput` variable always is a first param
```suggestion
        print_statistics(&mut output, &initial_statistics);
```

src/ips/algorithm.rs
```diff
@@ -128,11 +165,23 @@ impl Ips {
                 // after separation.
                 panic!("There are more than one massive island in the network. It is not possible to merge them automatically.");
```
Rqnsom, line 166: I haven't seen this change until now - nice!
(_this massive_count block check_)

src/ips/algorithm.rs
```diff
@@ -146,6 +195,8 @@ impl Ips {
 
         // Phase 2: Generate peer list using MCDA optimization.
 
+        writeln!(output, "MCDA is starting to work...").unwrap();
```
Rqnsom, line 198:
```suggestion
        writeln!(output, "The MCDA procedure is starting...").unwrap();
```

src/ips/statistics.rs
```diff
@@ -81,132 +82,220 @@ pub fn generate_statistics(state: &IpsState) -> Statistics {
...
-        "Median: {}",
-        stats.degree_median - stats_original.degree_median
-    );
-    println!(
-        "Min: {}, max: {}, delta: {}",
+/// Calculates percentage change between two values.
+fn percentage_change(original: f64, new: f64) -> f64 {
+    if original == 0.0 {
+        return 0.0;
+    }
```
Rqnsom, line 144: just a quick thought, not sure if this is a serious suggestion:
```rust
// Calc delta to keep the original value intact for this part
let delta = new - original;

if original == 0.0 {
    // We can use some small value here just to fake the infinity case (diving by zero)
    original = 0.00000001;
}

(delta / original) * 100.0
```

Rqnsom (approved):
LGTM with a few suggestions

src/ips/algorithm.rs
```diff
@@ -128,11 +165,23 @@ impl Ips {
                 // after separation.
                 panic!("There are more than one massive island in the network. It is not possible to merge them automatically.");
             }
+
+            writeln!(
+                output,
+                "IPS detected no massive islands however there are some disconnected nodes"
```
kylegranger, line 171: minor tweak: `IPS detected no massive islands. However, there are some disconnected nodes.`

testdata/sample_ips.log
kylegranger: hmm... I'll just mention, I prefer positive values without the `plus` sign.  Treat it is like any other numeric literal (IMO 😄 )

kylegranger (approved):
👍  The output is clear, detailed, and very readable.

testdata/sample_ips.log
Rqnsom: then let's keep it like it is - you will be using it more anyway probably 😃

src/ips/algorithm.rs
asmie: Done.

src/ips/algorithm.rs
asmie: Changed.

src/ips/algorithm.rs
asmie: Changed to `o` - feel better when it's `o` than `f` as this could suggest it's always some file.

src/ips/algorithm.rs
asmie: Changed.

testdata/sample_ips.log
asmie: Left as is for now 😃

PR #31, opened 2023-03-29
feat: add support for network type and num threads

kylegranger:


src/main.rs
```diff
@@ -56,10 +56,13 @@ async fn write_state(config: &CrunchyConfiguration) {
 
     geo_cache.configure_providers(&config.geoip_config);
 
+    const NUM_THREADS: usize = 8;
```
asmie, line 59: I wonder if there is any sense to have different value for that in different modules.
Because if not, maybe it could be better to put it in some constants.rs file and use it from there. It could also simplify API because we'd not pass it as an argument.

asmie (approved):
👍

src/main.rs
kylegranger: I like that idea a lot. Will do.
And, tbh, I've been thinking about the threading, and don't really like 'polluting' the spectre api with having to specify the thread count.  The teexgraph library I used just queried  the number of CPUs via the omp (C++) library:
```
   const int cpus = omp_get_num_procs();
```
I don't exactly know what the Rust equivalents might be, but I did see this:
https://docs.rs/num_cpus/latest/num_cpus/

So, in any event, I think the end game would be to remove `num_threads` throughout, and the compute call **just does it**.  Something we can discuss.

src/ips/algorithm.rs
```diff
@@ -465,8 +466,8 @@ impl Ips {
         let mut graph = construct_graph(nodes);
 
         if generate_full {
-            let betweenness = graph.betweenness_centrality();
-            let closeness = graph.closeness_centrality();
+            let betweenness = graph.betweenness_centrality(NUM_THREADS, false);
```
Rqnsom, line 469: general API comment:
it's impossible to say what `false` here means without checking the function documentation - so it would be good to replace it with an enum value

...5 minutes later, after I did my documentation check, here's a suggestion which would make it clear on the spot:
```rust
let betweenness = graph.betweenness_centrality(NUM_THREADS, Normalization::Enable);    // or NormalizationOpt::On
```

kylegranger: 1. Hmm, well I see clearly (at least in VS Code) that the the parameter is called `normalize`
Screenshot 2023-03-30 at 10 10 20


2. I move the mouse over the function name and get this:
Screenshot 2023-03-30 at 10 13 32
08.png)

3.  > it's impossible to say what false here means without checking the function documentation

Well, I would also say it's also impossible to use the function in your enum-instead-of-bool-case without checking the documentation.  Except in your case, to need to know an additional parameter type.

4.  Other public functions in the spectre API also use bools.

Screenshot 2023-03-30 at 10 25 47



5.  IMO, I don't see the point of using an enum type for `bool`, instead of just plain `bool`.  More text, more stuff to know, more complicated.  
6.  In my experience, the meaning of a boolean parameter is usually quite clear from the name. 
😄

kylegranger: I can fix this with next PR.  Since it's really a kind of constant -- just like any other **magic** number or string -- I'd define this in `constants.rs` (which was just added for the `NUM_THREADS` define).  I prefer a const name to an enum.
```
const NORMALIZATION_OFF: bool = false;
```

Rqnsom: > Hmm, well I see clearly (at least in VS Code) that the the parameter is called normalize

Okay, but that's the developer's perspective with VS Code - not everybody uses that IDE. The clean API approach shouldn't consider the IDE tools.

but the more important perspective:
I couldn't see anything like that while doing the code review on GitHub. I had to manually find the repo and seek the parameter there 🤷‍♂️

Rqnsom: >I can fix this with next PR. Since it's really a kind of constant -- just like any other magic number or string -- I'd define this in constants.rs (which was just added for the NUM_THREADS define). I prefer a const name to an enum.
>
>> const NORMALIZATION_OFF: bool = false;

it's another approach, but then the user of the API needs to ensure clean usage of the API - still not the cleanest solution, but it would definitely give much more context

PR #32, opened 2023-03-31
feat: network filtering

asmie:
Crunchy has now ability to filter network type that IPS will work on. This can be used to filter out not needed nodes which can belong to other networks.

Filtering is done over list of nodes and can be used in every place where one can use such structure. 

Filtering is implemented as a node removal - that seems to be shortest and easiest approach to the problem but probably not the optimal one (as we need to do more iterations removing not matching nodes that we could do selecting only matching nodes). However, removing nodes is simplier in terms of re-building all the connection info as it is just re-indexing indices higer than removing node. Re-creating whole connection graph seems to more complicated to track and re-numerate all other indices and connections.


zeapoz (approved):
LGTM :+1:

src/ips/graph_utils.rs
```diff
@@ -79,7 +80,23 @@ pub fn construct_graph(nodes: &[Node]) -> Graph<SocketAddr> {
...
+        // closeness centrality as well as simple getting degree.
+        if node.connections.is_empty() {
+            graph.insert(Edge::new(node_addr, node_addr));
+            continue;
+        }
+
         for i in &node.connections {
+            if *i >= nodes.len() {
+                println!(
+                    "Node {} has connection to non-existing node {}",
```
Rqnsom, line 95: when is this the case? because it sounds wrong - so maybe we can leave a comment with an example or something

src/ips/graph_utils.rs
```diff
@@ -79,7 +80,23 @@ pub fn construct_graph(nodes: &[Node]) -> Graph<SocketAddr> {
 
     for node in nodes {
         let node_addr = node.addr;
+
+        // This is a hack to add nodes that are not connected to any other node.
```
Rqnsom, line 84: Let's mention this can happen when certain nodes are found through a different network (e.g. not the most responsive zcash nodes discovered via the flux nodes)

src/ips/graph_utils.rs
```diff
@@ -90,8 +107,9 @@ pub fn construct_graph(nodes: &[Node]) -> Graph<SocketAddr> {
 /// Removes node from the state and updates all indices in the peerlist
 pub fn remove_node(nodes: &mut Vec<Node>, node_idx: usize) {
     let node = nodes[node_idx].clone();
-    for peer_idx in node.connections.iter() {
-        nodes[*peer_idx].connections.retain(|x| *x != node_idx);
+
+    for rnode in nodes.iter_mut() {
+        rnode.connections.retain(|x| *x != node_idx);
```
Rqnsom, line 112: would a combination of `position` and then `remove` be [faster](https://stackoverflow.com/a/44012406)?
```rust
// Ignore if no such element is found
if let Some(pos) = vec.iter().position(|x| *x == needle) {
    vec.remove(pos);
}
```

I tried to imagine a case where we have 1000 nodes, and we need to remove the first node - the `retain` solution would still go through all 1000 nodes, although the `node_idx==0`

src/ips/graph_utils.rs
```diff
@@ -123,6 +141,34 @@ pub fn find_lowest_betweenness(nodes_idx: &[usize], state: &IpsState) -> usize {
...
+                found = true;
+                break;
+            }
+        }
+
+        // If we didn't find any node with different network type we can stop.
+        if !found {
+            break;
+        }
+    }
```
Rqnsom, line 167: maybe an alternative approach (could be a bit faster):

in the first step - with a `for` loop, we find all non-zcash node indices (non-network_type)
```rust
        let mut bastard_indicies = Vec::with_capacity(network_nodes.len());

        // using `rev` so we don't mess up the order of those found indices once we start removing them
        for idx in (0..network_nodes.len()).rev() {          // could be done more idiomatically, probably
            if network_nodes[idx].network_type != network {
                bastard_indicies.push(idx);
            }
        }
```
in the second step - remove all those unwanted indices:
(using `reverse` so we don't mess up the order of those found indices)
```rust
    for undesired_idx in bastard_indicies {
        remove_node(&mut network_nodes, undesired_idx);
    }
```

Rqnsom (approved):
LGTM

src/ips/peer.rs
```diff
@@ -33,7 +33,7 @@ impl Peer {
         };
 
         for peer in &node.connections {
-            if *peer >= nodes.len() {
+            if *peer >= nodes.len() || nodes[*peer].addr == node.addr {
```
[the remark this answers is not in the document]
asmie: Yes, there are some paths that run peerlist generation right after re-constructing graph. As spectre does not allow to add single nodes I'm adding nodes connected to themselves. Then, if we'd like to generate peerlist it would contain such connections so this place get rid of them.

src/ips/graph_utils.rs
```diff
@@ -239,4 +285,73 @@ mod tests {
         assert_eq!(peers.len(), 1);
         assert!(peers.contains(&4));
     }
+
+    #[test]
+    fn filter_network_test() {
+        let nodes = vec![
+            Node {
```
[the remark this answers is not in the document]
asmie: 👍

src/ips/graph_utils.rs
asmie: Mentioned.

src/ips/graph_utils.rs
asmie: Changed.

src/ips/graph_utils.rs
asmie: Applied 😃

Rqnsom (approved):
🚀

PR #35, opened 2023-04-14
test: ability to generate file with original (vanilla) peerlist

asmie:
Feature for testing purposes.

That allows to look into peerlist right before IPS will run over it. Vanilla peerlist can be also used for visualization.
Samples updated.


kylegranger (approved):
LGTM 👍

Rqnsom (approved):
LGTM

PR #36, opened 2023-04-18
Add Node Filtering to Crunchy

kylegranger:
- optional filter parameter added (e.g., `Zcash`)
- filter the nodes, recompute node indices
- compute betweenness and closeness based on filtered graph.


src/nodes.rs
```diff
@@ -86,6 +86,110 @@ pub async fn create_nodes(
...
+    //   b) the index map
+    // We only keep connections where both nodes are in the index map
+    let mut graph = Graph::new();
+    for (n, node) in indices.iter().enumerate() {
+        let n_index: i32 = index_map[n];
+        if n_index != -1 {
+            node.iter()
+                .filter(|&connection| {
+                    // For each connection, we only add it once, so we use the connection
+                    // where source index is less than target
```
Rqnsom, line 127: nice comment!

src/nodes.rs
```diff
@@ -86,6 +86,110 @@ pub async fn create_nodes(
...
+}
+
+pub async fn create_nodes(
+    filter_type: Option<NetworkType>,
+    indices: &NodesIndices,
+    node_addrs: &[SocketAddr],
+    node_network_types: &[NetworkType],
+    geo_cache: &GeoIPCache,
+) -> Vec<Node> {
+    if filter_type.is_some() {
```
Rqnsom, line 179: more idiomatic appraoch:
```rust
if let Some(filter_type) = filter_type {
    create_nodes_filtered...
} else {
    create_nodes_unfiltered...
}
```

src/nodes.rs
```diff
@@ -86,6 +86,110 @@ pub async fn create_nodes(
...
+    // Create reindexing map using filter value
+    //    a) the nodes we keep get new indexing, 0..N
+    //    b) the nodes we don't want get -1
+    let mut index: i32 = 0;
+    for (n, network_type) in node_network_types.iter().enumerate() {
+        if network_type_filter == *network_type {
+            index_map[n] = index;
+            index += 1;
+        } else {
+            index_map[n] = -1;
```
Rqnsom, line 108: we can avoid this `else` block if we initialize all indices in the map to -1 from the start:
```
let mut index_map: Vec<i32> = vec![-1; num_nodes];
```

so the starting assumption is: we don't need the index_map index until we need it - but then we set it on the line 105

Rqnsom (approved):
with a quick look - LGTM

src/nodes.rs
kylegranger: ah, good idea... I missed that.

kylegranger: done

src/nodes.rs
kylegranger: done

Rqnsom (approved):

src/main.rs
```diff
@@ -113,6 +114,13 @@ async fn main() {
         configuration.ips_config.peer_file_path = arg_conf.ips_file;
     }
 
+    // Check if user error setting optional filter type
+    if arg_conf.filter_type.is_some() && arg_conf.filter_type.unwrap() == NetworkType::Invalid {
+        panic!("Invalid network type for filter. Should be Zcash or Ripple.")
```
asmie, line 119: Hopefully, when there will be new network added we won't forget to update this message :) Can be connected with previous review note and end up suggesting user reading the Readme to get possible values.

```suggestion
        panic!("Invalid network type for filter. Check Readme for possible values.")
```

src/nodes.rs
```diff
@@ -86,6 +86,108 @@ pub async fn create_nodes(
     nodes
 }
 
+pub async fn create_nodes_filtered(
```
asmie, line 89: It could be good to add even one test to check (especially for future) if it's working as expected.
I've checked that using below test (feel free to just copy&paste it).

```suggestion
#[cfg(test)]
mod tests {
    use crate::config::GeoIPConfiguration;
    use super::*;

    #[tokio::test]
    async fn create_nodes_filtered_test() {
        let indices = vec![vec![1, 2], vec![0, 2, 3], vec![0, 1, 3], vec![1, 2]];
        let node_addrs = vec![
            SocketAddr::from(([127, 0, 0, 1], 1234)),
            SocketAddr::from(([127, 0, 0, 2], 1234)),
            SocketAddr::from(([127, 0, 0, 3], 1234)),
            SocketAddr::from(([127, 0, 0, 4], 1234)),
        ];
        let node_network_types = vec![
            NetworkType::Unknown,
            NetworkType::Zcash,
            NetworkType::Unknown,
            NetworkType::Zcash,
        ];

        let config = GeoIPConfiguration::default();
        let mut geo_cache = GeoIPCache::new(&config);
        geo_cache.configure_providers(&config);

        let nodes = create_nodes_filtered(
            NetworkType::Zcash,
            &indices,
            &node_addrs,
            &node_network_types,
            &geo_cache,
        )
        .await;

        assert_eq!(nodes.len(), 2);
        assert_eq!(nodes[0].connections, vec![1]);
        assert_eq!(nodes[1].connections, vec![0]);

    }
}
```

src/main.rs
kylegranger: done!

src/nodes.rs
```diff
@@ -86,6 +86,108 @@ pub async fn create_nodes(
...
+            node_addrs,
+            node_network_types,
+            geo_cache,
+        )
+        .await
+    } else {
+        create_nodes_unfiltered(indices, node_addrs, node_network_types, geo_cache).await
+    }
+}
+
```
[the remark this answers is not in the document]
kylegranger: yes, good point.  
Thank you for the filtered test.
I've added two others, both using the 6103-node sample json (one filtered, one unfiltered)
Betweenness and closeness checks have been added.
An older (outdated) test has been removed.

README.md
```diff
@@ -61,6 +61,7 @@ Options:
   -o, --out-state <OUT_STATE>          Output file with state of the graph (overrides output from config file)
   -g, --geocache-file <GEOCACHE_FILE>  Output file with geolocation cache (overrides cache from config file)
   -c, --config-file <CONFIG_FILE>      Configuration file path (if none defaults will be assumed)
+  -f, --filter-type <FILTER_TYPE>      Optional node filtering by network type, e.g. 'Zcash'
```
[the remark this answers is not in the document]
kylegranger: done

asmie (approved):
Very nice! 👍

PR #37, opened 2023-04-24
Check and lint should be the same as in other repos

zeapoz:
The `check-and-lint` workflow in this repository was outdated and used deprecated actions. This PR adjusts it so it is similar to the others.


asmie (approved):
👍 
Now left wondering why `Cargo.lock` was initially added to .gitignore 😄 Good it's now as it should be 😃

PR #38, opened 2023-04-24
feat: `num_threads` argument

zeapoz:
This PR introduces:
- `num_threads` argument, that can be passed to control how many threads `spectre` computations will use. By default, ~~it will use as many as specified by `DEFAULT_NUM_THREADS`~~ it will guess the number of threads to use with `std::thread::available_parallelism`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -8,6 +8,8 @@ use crate::ips::config::IPSConfiguration;
 
 /// Default number of days to keep each entry in cache
 pub const DEFAULT_KEEP_IN_CACHE_DAYS: u16 = 14;
+/// Default number of threads to use
+pub const DEFAULT_NUM_THREADS: usize = 8;
 
 /// Main configuration structure
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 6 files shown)


src/config.rs
```diff
@@ -70,6 +74,7 @@ impl Default for CrunchyConfiguration {
             ips_config: IPSConfiguration::default(),
             geoip_config: GeoIPConfiguration::default(),
             network_type_filter: None,
+            num_threads: DEFAULT_NUM_THREADS,
```
kylegranger, line 77: What do you think of the idea of using `num_cpus::get()` or `std:thread:available_parallelism` to obtain, perhaps, a better default number-of-threads?  This wasn't part of original task, but it might be the right thing to do.  This had been my original thought for `spectre`, but I've amended my thinking on that:  I think it's good that the caller has control over how many threads are used.
Btw, the changes look great! 👍

Rqnsom: > but I've amended my thinking on that: I think it's good that the caller has control over how many threads are used.

This is how I think here: I would probably always use the max number of threads without thinking much - but I'm wondering, when would one decide to use less than max threads?

I don't mind either option tbh, I'm just thinking at loud

asmie: When eg. you're running crunchy locally and would like to do something more than just looking how crunchy consumes your CPU 😄

kylegranger (approved):
very nice! 👍

